Issue #97, opened 2026-02-11
Create automated region selection for evaluation mode


o-ostrovskiy:
## Summary

StoryLand workflow requires human interaction to select a region from the discovery results. For automated evaluation, we need a deterministic region selection strategy.

## Current Behavior

In `main.py`:
- `display_region_options()` shows regions to user
- `get_region_selection()` waits for user input
- User selects which region(s) to explore

This blocks automated evaluation.

## Proposed Solution

Add an `evaluation_mode` flag that enables automatic region selection:

**Option 1: Always select first region**
```python
if evaluation_mode:
    selected_regions = [regions[0]]  # Deterministic
else:
    selected_regions = get_region_selection(region_analysis)
```

**Option 2: Select region with most cities**
```python
if evaluation_mode:
    selected_regions = [max(regions, key=lambda r: len(r['cities']))]
```

**Option 3: Configurable via test case metadata**
- Allow eval cases to specify which region to select
- Falls back to first region if not specified

## Related Files

- `main.py` (lines 78-110 - region selection logic)
- `tools/run_scheduled_eval.py` (needs to pass evaluation_mode flag)
- `agents/orchestrator.py` (workflow configuration)

## Acceptance Criteria

- [ ] Evaluation mode bypasses human region selection
- [ ] Selection strategy is deterministic and documented
- [ ] Optional: Support configurable selection per test case
- [ ] Workflow completes end-to-end without user input
- [ ] Tests verify automated selection works

PR #130, opened 2026-03-19
129 New Evaluation Dataset with Quality Criteria Injection into LLM J…

SvetlanaNMik:
Created evaluation/books_v1.evalset.json — 10 cases covering Lord of the Rings, Stephen King, The Expanse, Red, White & Royal Blue, García Márquez, The Kite Runner, Dostoevsky, Cheryl Strayed, Dracula, and Shogun
Each case includes query, user_preferences, expected_output, and quality_criteria
Registered as books_v1 dataset in Langfuse
llm_scorer.py — _build_scoring_prompt() now accepts quality_criteria (per-dimension book-specific requirements appended as "Book-specific requirement:" lines) and expected_output (rendered as a **REFERENCE OUTPUT** section for the judge to compare against)
run_scheduled_eval.py — extracts quality_criteria and expected_output from Langfuse item metadata and passes both to score_itinerary()
langfuse_eval.py — stores quality_criteria in Langfuse item metadata at registration time; validates keys against known 6 dimensions and warns on unknown/misspelled keys
langfuse_eval.py — added --evalset-file CLI flag to register one evalset at a time
Makefile — added eval-setup-one EVALSET_FILE=<path> target

### DIFF
--- a/evaluation/tools/llm_scorer.py
+++ b/evaluation/tools/llm_scorer.py
@@ -198,11 +198,8 @@ async def score_itinerary(
     input_text: str,
     itinerary: Dict[str, Any],
     preferences: Optional[Dict[str, Any]] = None,
-<<<<<<< 129-evaluation-pipeline-enhancements
     quality_criteria: Optional[Dict[str, str]] = None,
     expected_output: Optional[Dict[str, Any]] = None,
-=======
->>>>>>> main
     model_name: str = "gemini-2.5-flash",
     langfuse_trace_id: Optional[str] = None,
     langfuse_client: Optional[Any] = None,
--- a/evaluation/tools/run_scheduled_eval.py
+++ b/evaluation/tools/run_scheduled_eval.py
@@ -657,11 +657,8 @@ Include ALL cities from the selected regions in your itinerary."""
                     input_text=book_title,
                     itinerary=itinerary_data,
                     preferences=preferences,
-<<<<<<< 129-evaluation-pipeline-enhancements
                     quality_criteria=quality_criteria,
                     expected_output=expected_output,
-=======
->>>>>>> main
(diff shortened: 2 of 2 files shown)


o-ostrovskiy:
**🔴 P1-B — Single-region runner vs. multi-region `expected_output` makes scores systematically invalid**

`select_first_region()` (issue #97) always picks one region for deterministic evaluation, but several `expected_output` entries in `books_v1` describe full multi-region itineraries — query_011 covers NZ + England + Ireland, query_014 covers London + DC + NYC, query_016 and query_019 are similarly multi-destination.

The LLM judge now receives these as benchmarks and penalises every output for "missing" destinations the runner is structurally forbidden from including. `completeness` and `book_relevance` scores will be systematically depressed and not comparable across runs. This isn't a flaky test — the bias is deterministic and permanent until fixed.

**Options (in order of preference):**

1. **Scope `expected_output` to single-region** — update the four affected cases so the reference describes a strong single-region itinerary the runner can actually match. The benchmark should cover the system's reachable output space. This is the cleanest fix.

2. **Null out `expected_output` for known multi-region cases** — in `_run_evaluation_case`, pass `expected_output=None` to `score_itinerary` when the reference spans multiple distinct geographic regions. Fragile but unblocks the PR.

3. **Implement multi-region composition** — change `select_first_region` to select all regions. Right long-term answer, but out of scope here.

Whichever path is taken, please add a note to `evaluation/README.md` that cross-run score comparisons are only valid when the same region-selection strategy is used — otherwise historical scores from the single-region era won't be comparable once multi-region lands.

SvetlanaNMik:
@o-ostrovskiy 

- books_v1 with 10 evals registered in Langfuse

- select_all_regions() implemented

- All traces are now visible and populated.

image

PR #139, opened 2026-04-08
Refine agent prompts based on evaluation results

### LINKED ISSUE #63: Improve prompts based on eval feedback
## Description
Refine agent prompts based on evaluation results.

## Tasks
- [ ] Identify low-scoring areas
- [ ] Update agent prompts
- [ ] A/B test prompt variations in Langfuse
- [ ] Document prompt versions

**Sprint:** S10 (Jun) | **Hours:** 5

SvetlanaNMik:
A/B test results — books_v1 (10 cases, controlled runs)

Two full runs on the same dataset — one with v1 prompts, one with v2 — showed improvement across all 6 evaluation dimensions. The biggest gains were in completeness (+0.4) and actionability (+0.3), the two dimensions most directly targeted by the prompt changes. preference_adherence improved by +0.2 and book_relevance by +0.3. geographical_accuracy and engagement each improved by +0.2.

The largest per-book gain was Leviathan Wakes, which jumped from 1.0 to 3.67. This book was a recurring all-1s failure due to the city_formatter hallucinating fictional space locations as real cities — eliminated by the country validation gate across 4 consecutive runs.


agents/discovery_agents.py
```diff
@@ -30,13 +30,27 @@ def create_city_pipeline(model, google_search_tool):
...
 - "[book title] [author name] inspiration places"
 - "[setting location] literary tourism"
 
+For recent fiction (published after 2000) that may not yet have established literary tourism, also try:
+- "[city] [book title] filming location"
+- "[book title] real-life inspiration"
+- "[author name] [city] residence"
+Even a loose connection (author lived there, setting inspired by the city) is valid.
+
+**FICTIONAL/SPACE/FUTURE WORLDS — MANDATORY REDIRECT:** If the book's primary_locations from context are clearly non-Earth (space stations, fictional planets, asteroid belts, alien worlds, invented continents), you MUST NOT search for those locations and you MUST NOT return empty-handed. Instead, call google_search immediately with these queries and return the results:
```
o-ostrovskiy: I like it :-)

evaluation/tools/run_scheduled_eval.py
```diff
@@ -112,6 +112,7 @@ async def run_evaluation_on_dataset(
     item_ids: Optional[List[str]] = None,
     start_time: Optional[float] = None,
     timeout_seconds: Optional[float] = None,
+    prompt_version: str = "v2",
```
o-ostrovskiy: @SvetlanaNMik Both run_evaluation_on_dataset() and run_all_evaluations() have prompt_version: str = "v2" as default arguments. 
This will silently mis-tag Langfuse runs as v2 whenever these functions are called programmatically without explicitly passing the version — which is exactly what will happen after v3 lands. 
The CLI default (--prompt-version v2) has the same problem. 

Consider extracting a module-level constant:
CURRENT_PROMPT_VERSION = "v2"

hen use it in all three defaults:

prompt_version: str = CURRENT_PROMPT_VERSION
parser.add_argument('--prompt-version', default=CURRENT_PROMPT_VERSION, ...)

A single update point prevents silent version drift in evaluation data.

evaluation/README.md
```diff
@@ -90,6 +90,10 @@ Evaluations run automatically via GitHub Actions:
 
 Manual trigger: Go to Actions tab → Scheduled Evaluation → Run workflow
 
+### Prompt Versioning
+
+Use `--prompt-version <label>` to tag a run in Langfuse before merging a prompt change. Compare against previous runs by filtering on `prompt_version` in the Langfuse Runs view. Prompt change history is in [`evaluation/PROMPT_CHANGELOG.md`](PROMPT_CHANGELOG.md).
```
o-ostrovskiy, line 95: @SvetlanaNMik Great job! Really like this prompt version feature

agents/discovery_agents.py
```diff
@@ -115,14 +131,16 @@ def create_landmark_pipeline(model, google_search_tool):
 - Which city it's in
 - Specific connection to the book (Was it mentioned? Does it relate to a scene?)
 
+For non-urban books (trails, wilderness, road trips), scenic and route-based stops are valid: trailheads, viewpoints, rest stops, and local cafés along the route. Label these with confidence level "scenic_stop" or "route_point".
```
o-ostrovskiy: models/discovery.py still only preserves name, city, and connection for each landmark. That means the new labels are dropped before composition, and the required city field pushes wilderness stops like trailheads/viewpoints/rest stops toward either hallucinated towns or formatter rejection.

o-ostrovskiy: Layer 1 - The label is never captured by the formatter. The landmark_researcher is now told to label stops with confidence_level "scenic_stop" or "route_point", but the landmark_formatter instruction (unchanged in this PR) maps research output into LandmarkInfo with exactly three fields: name, city, connection. There is no type or confidence_level field in the instruction or the schema, so the label is silently discarded.
Layer 2 — The city field is required and blocks valid wilderness stops. LandmarkInfo.city: str is not optional, and the formatter instruction explicitly validates: "each landmark has a name, city, and connection — skip landmarks with insufficient information." A trailhead, scenic overlook, or rest stop on the Pacific Crest Trail has no city. The formatter will either reject it outright or the LLM will hallucinate the nearest town to satisfy the validation gate — the exact same class of hallucination this PR was trying to fix for fictional worlds.
Layer 3 — Even if data survived, the trip composer can't distinguish stop types. CityStop.type in models/itinerary.py is a free-form string, so scenic_stop/route_point would be valid there. But the composer reads from landmark_discovery in conversation history, and by that point the type is already gone. The composer prompt instructs it to use scenic_stop and route_point — but it has no signal from the discovery data telling it which stops warrant those types.
What the fix looks like
The model needs to grow to match the prompt's intent:
python# models/discovery.py
class LandmarkInfo(BaseModel):
    name: str = Field(description="Landmark or place name")
    city: Optional[str] = Field(
        default=None,
        description="City where landmark is located. May be null for non-urban stops (trailheads, scenic overlooks, route stops)."
    )
    region: Optional[str] = Field(
        default=None,
        description="Region or area for non-urban stops (e.g., 'Sierra Nevada, California')"
    )
    connection: str = Field(description="How this landmark relates to the book")
    landmark_type: Optional[str] = Field(
        default=None,
        description="Type of stop: 'mentioned in book', 'related to setting', 'suggested for atmosphere', 'scenic_stop', 'route_point'"
    )
The landmark_formatter instruction also needs to be updated to map scenic_stop/route_point labels into landmark_type, allow null city for non-urban stops, and use region as a fallback. And the formatter's validation gate ("skip landmarks with insufficient information") needs a carve-out: a stop with name + region + connection is sufficient when city is null.

agents/discovery_agents.py
```diff
@@ -72,6 +86,7 @@ def create_city_pipeline(model, google_search_tool):
 - Accept partial results (1-2 cities) if that's all that was found
 - Validate that each city has at minimum a name and country
 - Skip cities with insufficient information rather than guessing
+- VALIDATION GATE: The `country` field MUST be a real, sovereign nation that exists on Earth today (e.g. "Romania", "USA", "Japan"). Reject ANY entry where the country is a planet ("Mars"), moon, asteroid, space station, fictional place, "Solar System", "Outer Space", or anything that is not a real Earth country. If no valid Earth-country cities were found, return an empty cities list.
```
o-ostrovskiy: The gate as written:

"The country field MUST be a real, sovereign nation that exists on Earth today (e.g. 'Romania', 'USA', 'Japan'). Reject ANY entry where the country is a planet…"

The word "sovereign nation" is doing the wrong job here. It's a political science term, not a travel term, and LLMs know the difference. The gate was designed to block "Mars" and "The Asteroid Belt", but the test it actually runs — is this a sovereign state? — would also fail on:

England, Scotland, Wales — constituent countries of the UK, explicitly not sovereign nations. This is a problem right now: Dracula has locations in both Transylvania and Whitby/London. If the LLM formats those as "England" (which it commonly does, not "United Kingdom"), the gate rejects them.
Hong Kong — SAR of China, huge literary tourism destination for dozens of novels.
Puerto Rico, Guam — US territories with no sovereignty.
Catalonia, Bavaria, Provence — regions often returned when a book is set there rather than in "Spain" / "Germany" / "France".
Palestine, Taiwan, Kosovo — recognized by some countries, not others; an LLM reasoning about "sovereign nations" may exclude these inconsistently.

The core problem is that the intent of the gate (block non-Earth fictional locations) and its implementation (must be a sovereign nation) are different things. The fix is to rewrite it around the actual exclusion criteria:
diff- VALIDATION GATE: The `country` field MUST be a real, sovereign nation that exists on
-   Earth today (e.g. "Romania", "USA", "Japan"). Reject ANY entry where the country is a
-   planet ("Mars"), moon, asteroid, space station, fictional place, "Solar System",
-   "Outer Space", or anything that is not a real Earth country. If no valid Earth-country
-   cities were found, return an empty cities list.

+ VALIDATION GATE: The `country` field must refer to a real, Earth-based location that
+   a traveller can visit today. Sovereign nations ("France", "Japan"), constituent
+   countries ("England", "Scotland", "Wales"), territories and SARs ("Hong Kong",
+   "Puerto Rico"), and commonly used regional names ("Catalonia", "Provence") are all
+   valid. Reject ONLY entries where the country is clearly non-Earth: a planet ("Mars"),
+   moon, asteroid, space station, fictional world, or invented place ("The Outer Rim",
+   "Arrakis", "Solar System"). If no Earth-based cities remain after filtering,
+   return an empty cities list.
The revised wording keeps the Leviathan Wakes fix intact — "Mars" and "The Asteroid Belt" still get blocked — while not silently wiping Edinburgh, Cardiff, or Hong Kong from results. It also explains the rule in terms of what a traveller can actually visit, which is the right frame for this application.
One secondary note: the city_researcher prompt also has an example "- country: Country name (e.g., 'France')" in its formatter instructions. That example should add "England" or "Scotland" alongside "France" so the LLM's prior toward sovereign-nation formatting is nudged the right direction before the validation gate even runs.

o-ostrovskiy (changes requested):

SvetlanaNMik:
All agent instruction strings were hardcoded inside Python factory functions. This made it impossible to switch prompt versions at runtime — the --prompt-version flag in the eval runner only added a Langfuse metadata label; it did not actually change which prompts were used. 
Moved all 11 agent instruction strings to versioned JSON files in agents/prompts/. Each file is a complete, standalone snapshot of the full prompt set for that version. A new agents/prompts.py module provides an AgentPrompts dataclass and a load_prompts(version) function with in-process caching.

All agent factory functions and the orchestrator now accept an optional prompts: AgentPrompts | None = None parameter. When omitted, the default version (v2) is loaded automatically — so main.py and any existing callers require no changes. The eval runner's --prompt-version flag now loads and applies the matching JSON file, making A/B comparisons fully automated.

Files:
agents/prompts/v2.json  New — current prompt set (all 11 agents)
agents/prompts/v1.json  New — original prompts 
agents/prompts.py	New — AgentPrompts dataclass + load_prompts() with cache
agents/trip_composer_agent.py	Uses prompts.trip_composer from loaded JSON
agents/discovery_agents.py	Uses per-agent keys from loaded JSON
agents/book_context_agent.py	Uses JSON template + .format(book_ref, search_hint)
agents/reader_profile_agent.py	Uses prompts.reader_profile from loaded JSON
agents/region_analyzer_agent.py	Uses prompts.region_analyzer from loaded JSON
agents/orchestrator.py	Threads prompts through to all factory calls
evaluation/tools/run_scheduled_eval.py	Calls load_prompts(prompt_version) and passes to workflows
README.md	Updated project structure tree and Prompt Engineering section

# Run eval with v1 prompts (original):
python evaluation/tools/run_scheduled_eval.py --dataset single_test --max-cases 1 --prompt-version v1

# Run eval with v2 prompts (current default):
python evaluation/tools/run_scheduled_eval.py --dataset single_test --max-cases 1 --prompt-version v2

# Add a new version: create agents/prompts/v3.json, then:
python evaluation/tools/run_scheduled_eval.py --prompt-version v3

SvetlanaNMik:
image

o-ostrovskiy (approved):
Great job!!!

Issue #15, opened 2026-01-28
Add Redis caching for book metadata


o-ostrovskiy:
## Description
Cache Google Books API responses to reduce API calls.

## Tasks
- [ ] Set up Redis client
- [ ] Cache book metadata with TTL (24 hours)
- [ ] Add cache hit/miss logging
- [ ] Implement cache invalidation

**Sprint:** S8 (May) | **Hours:** 4

Issue #16, opened 2026-01-28
Create OpenAPI/Swagger documentation


o-ostrovskiy:
## Description
Generate comprehensive API documentation.

## Tasks
- [ ] Add docstrings to all endpoints
- [ ] Configure Swagger UI
- [ ] Add example requests/responses
- [ ] Document error codes

**Sprint:** S8 (May) | **Hours:** 3

o-ostrovskiy:
OpenAPI/Swagger is configured via FastAPI metadata in api/app.py, with endpoint docstrings/examples in api/routes.py and typed schemas in api/models.py.

PR #207, opened 2026-07-19
refactor(core): shared run harness — PR 1 of the ADK 2 migration

o-ostrovskiy:
## Summary

First PR of the planned ADK 1.x → 2.x migration (behavior-frozen, still on ADK 1.36.1). Extracts the ADK-facing scaffolding that was copy-pasted across the five `WorkflowExecutor` flows and the place→book resolver into `core/run_harness.py`, so the upcoming ADK bump changes **one** code path instead of six.

- **`pump_events`** — the single drain loop for `runner.run_async`: per-agent progress mapping, researcher-text capture, and final-response tracking via `RunCapture`.
- **`run_guarded` + `GuardSpec`** — the workflow timeout plus the Timeout/Cancel/Exception boundary. Per-flow cleanup policy is preserved exactly as data (e.g. `expand`'s exception path clears the lock but does **not** mark the session failed, while its timeout path does both). `discover`'s TaskGroup-classification keeps its own `map_exception` hook.
- **`collect_token_usage` / `error_events`** — token close-out and the standard `error → done` terminal pair.
- Dedupe: `_normalize_text` now lives once in `core/extraction.py`; the two grounding *filters* stay separate deliberately (different fail-open contracts, documented in place).

## Behavior freeze

- All **727 pre-existing unit tests pass unmodified** — Runner construction stays at the historical module-level seams (`core.executor.Runner`, `core.place_to_book.Runner`), so no monkeypatch target moved.
- New **golden-stream test** snapshots the exact SSE event sequence for discover→compose (happy path + rejection paths) through the real executor/harness/`domain_event_to_sse` seam. This is the wire-contract gate the ADK 2 PRs must keep green.
- 17 new tests total; **744 passed**, integration suite green, coverage **82.56%** (ratchet floor 74%).

## Docs

ADR #15 (docs/ARCHITECTURE.md) records the design and trade-offs; README tree + stale test count (442 → 744) updated.

## Next in the stack

PR 2: ADK 2.4 lift (deps/CI, Langfuse plugin port to the public `branch` API, session layer + retention rewrite) — spike first, evals gated against the pre-migration baseline.


o-ostrovskiy:
**Reviewed @a64b573142566f82ee2ed6702385f97698ab98da — held on a conflict, not on the code.**

`mergeable_state: **dirty**`. This head hasn't moved since 12:45 while `main` took two merges since: ai#213 (`4fc46d9`, `reader_profile_agent` removed — touches `core/executor.py`'s step maps, which this PR rewrites wholesale) and ai#212 (`5688fcd5`, the decision-log backfill — touches `docs/ARCHITECTURE.md`, `README.md`, `api/routes.py`, `core/place_to_book.py`).

**Resolve by merging `origin/main` INTO `refactor/run-harness`** — a merge commit, never rebase/force-push (org convention; a force-push here would also break #209, which is stacked on this branch). Keep both sides' additive hunks and re-run the suite.

**Two specific collisions to expect, both mine to have caused:**

1. **`core/executor.py`** — ai#213 deleted the `"reader_profile_agent"` entries from `DISCOVERY_AGENT_STEPS` and `LOCAL_ATMOSPHERE_AGENT_STEPS`. This PR restructures the same file. Take main's deletion; the agent is gone from both workflows and must not come back through the harness extraction. There's a class-level guard on main (`TestNoReaderProfileAgent`) that will red if it does — good, let it.
2. **`docs/ARCHITECTURE.md`** — this PR documents the harness as **ADR #15**, but ai#212 merged an hour ago and took #15 (persistent cache) and #16 (`place_key`). Renumber to the next free entry and repoint references. I created that collision by merging #212 while this stack was held without checking its ADR numbers — flagging it rather than letting it surface as a merge artifact.

**On the code itself:** unreviewed at full depth pending the conflict resolution, since the resolved tree is what I'd be signing. What I can say from the first pass stands — the behaviour-freeze evidence is the right evidence (727 pre-existing tests unmodified, Runner construction left at the historical module-level monkeypatch seams, and a golden-stream SSE snapshot as the wire-contract gate the ADK 2 PRs must keep green). Extracting one path before changing it is the correct order of operations for this migration.

**Merge method — deliberate deviation, stated up front:** I intend to merge this with a **merge commit** rather than our standard squash. Squashing would put the harness changes on `main` as one new commit while #209 still carries the originals, guaranteeing conflicts across exactly the files ADK 2 rewrites. Preserving history here makes the stacked PR merge cleanly. Knowingly departing from convention, for that reason.

→ In Progress, `Role: Staff Engineer`. Re-review on the resolved head; #209's own blocker is [on that PR](https://github.com/o-ostrovskiy/storyland-ai/pull/209#issuecomment-5017052422).

Issue #23, opened 2026-01-28
Create 10 more eval queries (30 total)


o-ostrovskiy:
## Description
Expand evaluation dataset to 30 books.

**Sprint:** S12 (Jul) | **Hours:** 3

Issue #22, opened 2026-01-28
Add cost tracking dashboard in Langfuse


o-ostrovskiy:
## Description
Track and visualize Gemini API costs.

## Tasks
- [ ] Calculate cost per request
- [ ] Create daily/weekly cost reports
- [ ] Set up cost alerts

**Sprint:** S11 (Jul) | **Hours:** 3

Issue #17, opened 2026-01-28
Build quality scoring metrics in Langfuse


o-ostrovskiy:
## Description
Create automated quality scoring for generated itineraries.

## Metrics
- Has literary sites (boolean)
- Has author connections (boolean)
- Days within reasonable range
- Geographic coherence
- Response latency

**Sprint:** S9 (Jun) | **Hours:** 5

PR #209, opened 2026-07-19
feat(deps): google-adk 2.x lift, template-first — PR 2 of the ADK 2 migration

o-ostrovskiy:
## Summary

Stacked on #207. Bumps `google-adk[eval]>=1.33,<2` → **`google-adk[db]>=2.4.0,<3`** (locks resolve 2.5.0) and adapts every 2.x-breaking surface, deliberately keeping the deprecated-but-functional `SequentialAgent`/`ParallelAgent` template workflows so the dependency jump and the graph-runtime rewrite (PR 3) are separately bisectable. The model stays pinned to `gemini-2.5-flash-lite` so any eval delta is attributable to ADK alone.

### Spike-verified fixes (against the real 2.4/2.5 package, isolated venv first)

- **Gemini auth**: `api_key=` is *silently dropped* on 2.x (pydantic ignores it; auth falls back to env — a dangerous misconfiguration, not an error). Both construction sites now pass `client_kwargs={"api_key": ...}`.
- **Langfuse plugin**: `_branch_key` uses the now-public `CallbackContext.branch` property, retiring the private `_invocation_context` coupling that motivated the old `<2` pin (private path kept as fallback only). Missing `usage_metadata` now logs a WARNING instead of silently zeroing cost tracking.
- **Sessions**: fresh-DB cutover — default file renamed `storyland_sessions.db` → `storyland_sessions_v2.db` so 2.x never opens a 1.x-schema file. `[db]` extra restores sqlalchemy (no longer a core dep); explicit `greenlet` for the async engine.
- **Retention sweeper**: verified compatible (v-current schema keeps `sessions.update_time`; engine attr is `db_engine`) and now **proven** by a new integration test that drives a real ADK 2 SQLite store and asserts stale-session eviction — the prune was fail-open and untested before.
- **Security**: the five starlette PYSEC ignores (2026-161/248/249/2280/2281) are **removed** from `codex.yml` + `Makefile` — ADK 2.x lifts starlette to 1.3.1 which carries all the fixes. `make audit` passes with only the fixless diskcache ignore. CI pin guard flipped to assert a 2.x resolution; `[eval]` extra dropped (nothing imports `google.adk.evaluation`).
- `tools/preferences.py`: documented the deliberate fail-open (2.x only auto-retries a tool whose exception propagates).

### Verification

- **744 unit tests pass** — 743 passed unmodified on ADK 2.5.0 on the first run (the PR 1 harness paid off); the only change was flipping the intentional version-guard test from 1.x to 2.x.
- Integration gate green, including the new retention-sweep-vs-real-store tests.
- **Golden-stream SSE snapshot unchanged** — the wire contract survives the major bump.
- Coverage 82.56% (floor 74). Lockfiles regenerated (`make lock`), CI staleness gate satisfied.

### Docs

ADR #16 records the migration (decisions + spike findings); ADR #4 rewritten to match the actual bounded-backoff retry code (it had documented the old inverted `exp_base=7` schedule since June); AGENTS.md/claude.md dependency guidance updated; README session-schema note.

### Deploy note

First deploy of this PR is the fresh-DB cutover: update `.env.prod` `DATABASE_URL` to the `_v2` filename, deploy in a quiet window (in-flight jobs at that moment are lost — accepted), keep the old DB file until PR 4 stabilizes, then delete.


o-ostrovskiy:
**Eval gate for this PR** (dispatched on this branch: [run 29688023901](https://github.com/o-ostrovskiy/storyland-ai/actions/runs/29688023901), same model `gemini-2.5-flash-lite`):

Baseline = two runs on `main` @ 4034b5a:
| dataset | run1 | run2 | pooled | measured noise |
|---|---|---|---|---|
| storyland_eval (avg/5) | 4.42 (n=8) | 4.25 (n=8) | **≈4.33** | ±0.17 |
| books_v1 (avg/5) | 3.33 (n=10) | 2.93 (n=9) | **≈3.14** | ±0.40 (12%) |

Because books_v1's run-to-run noise exceeds a flat −5% threshold, the gate is calibrated to measured noise: **pass = post-change avg ≥ pooled − 0.35 (storyland_eval) and ≥ pooled − 0.50 (books_v1)**; a miss triggers one rerun before being treated as a real regression. (The workflow's overall "failure" status is the pre-existing `place_to_book_v1` dataset-format mismatch — judge scores in the artifacts are the signal.)

o-ostrovskiy:
**Eval gate result — PASS** ([run 29688023901](https://github.com/o-ostrovskiy/storyland-ai/actions/runs/29688023901), this branch, ADK 2.5.0, `gemini-2.5-flash-lite`):

| dataset | this branch | pooled baseline | gate | verdict |
|---|---|---|---|---|
| storyland_eval | **4.33** (n=8) | 4.33 | ≥ 3.98 | ✅ PASS (dead-on baseline) |
| books_v1 | **3.20** (n=10) | 3.14 | ≥ 2.64 | ✅ PASS (above baseline) |

Tokens/cost per case: ~45.8k / $0.0070 (storyland), ~44.6k / $0.0068 (books) — marginally *below* the 1.x baseline. No quality or cost regression from the ADK 1.36 → 2.5 jump.

o-ostrovskiy:
**Reviewed @56553754ed48e4679a4078ed507953d40751c61a — not merging yet. One substantive item, three mechanical.**

The eval gate clearing is the right outcome and the calibration work behind it (two-run baseline, measured judge noise, noise-aware thresholds replacing the flat −5%) is the correct way to build that gate. This is not a re-litigation of the decision to land.

---

### 🔴 1. The instrument that produced the cost numbers has a known blind spot — and this PR is where it was found

Codex's `usage_metadata is None` finding is still open on this head, and I've now confirmed it at source rather than from the summary:

```python
if hasattr(response, 'usage_metadata'):
    metadata = response.usage_metadata          # may be None
    return TokenUsage(
        input_tokens=getattr(metadata, 'prompt_token_count', 0) or 0,
        ...
    )
```

`LlmResponse` is a pydantic model, so `usage_metadata` is a **declared field** — `hasattr` is always `True`, even when the value is `None`. `getattr(None, 'prompt_token_count', 0)` → `0`. The function therefore returns `TokenUsage(0, 0, 0)`, which is a plain dataclass with no `__bool__`, so **`if usage:` is truthy**. Result: the success branch runs, `generation.update()` records `usage_details {0,0,0}` and `cost_details {total_cost: 0.0}`, and your new `langfuse_token_usage_missing` WARNING — the one added precisely to make this loud — never fires. A silent zero-cost generation, in Langfuse, indistinguishable from a real one.

**Why this blocks rather than rides along:** you cleared the gate partly on *"tokens and cost per case came out marginally lower than the 1.x baseline."* Marginally-lower aggregate cost is the exact signature this defect produces. I am not claiming the numbers are wrong — the same code shape exists on the 1.x line, so if `usage_metadata` is never `None` on either side, both measurements are sound and the comparison holds. But the whole reason this PR added the missing-usage warning is the suspicion that a major ADK version could change that, and it's the one hypothesis the current instrument cannot report on.

**What I want, and it costs no spend:** don't re-run the eval. Query the Langfuse traces already captured from both eval runs for generations with `total == 0` tokens. Zero such generations on both sides → the comparison stands as measured, and the code fix below is pure robustness. Any such generations on the 2.5 side → the per-case cost is understated by construction and the cost half of the gate needs re-reading before it means anything.

**Fix regardless:** treat `metadata is None` as missing before constructing `TokenUsage`, so the None case falls to the `else` and warns. One condition, plus a falsifier asserting the WARNING fires for a response whose `usage_metadata` is `None` (a test that reds against this head).

---

### 2. Mechanical: this PR's `mergeable_state: clean` is meaningless right now

Its base is `refactor/run-harness`, not `main` — clean against a base that is itself conflicted. #207 currently reads **`dirty`**. Nothing here can merge until #207 lands.

### 3. 🔴 ADR numbering now collides on `main` — my fault, not yours

I merged ai#212 (the decision-log backfill) an hour ago. It took **ADR #15 = persistent content-versioned cache** and **ADR #16 = cross-job `place_key`**. #207 documents its harness as ADR #15 and this PR documents the migration as ADR #16. Both are now taken. I made that collision by merging #212 first without checking the held stack's ADR numbers — worth saying plainly since this PR's own review round is what taught us that a decision log must not claim an unmerged decision.

Renumber both to the next free entries (#22, #23) and repoint the cross-references. Purely clerical, but if it lands wrong the log is wrong on day one of §6.

### 4. Merge mechanics for the stack — a deliberate convention deviation, flagging before I do it

Our standing method is squash. If I squash #207, `main` gets one new commit containing the harness changes while this branch still carries the *original* harness commits, and merging `main` in afterwards will conflict across exactly the files ADK 2 also rewrote. I intend to merge **#207 with a merge commit** so its history exists on `main` and this branch merges cleanly. Deviating from the squash convention knowingly and for a stated reason; say if you'd rather I didn't.

---

**Order once the above is done:** resolve #207's conflict (merge `origin/main` **in**, no rebase) → I review and merge #207 → retarget this PR to `main`, merge `main` in → re-verify CI + the golden-stream SSE snapshot on the new head → I merge this. The `.env.prod` `_v2` cutover then rides the next ai deploy in a quiet window, on Olga's G5.

o-ostrovskiy:
**Codex finding: confirmed, fixed, and the gate conclusion audited against it.**

**Audit of the shipped numbers** (from the CI logs of the three eval runs already captured — no re-run):

| run | side | LLM responses | langfuse generation lines | zero-token generations | independent LoggingPlugin usage lines (zero) |
|---|---|---|---|---|---|
| 29687220782 | 1.x baseline | 216 | 216 | **0** | 216 (0) |
| 29687614009 | 1.x baseline | 216 | 216 | **0** | 216 (0) |
| 29688023901 | 2.5 gate | 216 | 216 | **0** | 216 (0) |

`usage_metadata` was never None on either side, so the cost half of the gate comparison stands as measured. (Same 216-generation count on both majors is itself a nice topology invariant.)

**The fix** (5fa649d): `_extract_token_usage` guards the None value behind the declared-field `hasattr` trap, normalizes an all-zero extraction to None (a real generation can never be 0/0/0), and the caller checks `usage is not None` — so the `langfuse_token_usage_missing` warning now owns every missing-usage response instead of being dead code. 8 regression tests pin the contract.

Also in this push: merged the updated base (main's ADR backfill + reader_profile removal come through), migration ADR renumbered **#16 → #23** with cross-refs fixed to the harness ADR's new **#22**.

o-ostrovskiy:
**Critique @5fa649d30d6d722c921dbf35cb5ae93de42188d2 — the blocker I raised is properly closed. Six criticisms, none of them a hard stop, two I'd defend hard.**

The audit was the right move: two independent instruments, 216/216 on both sides, no re-run, no spend. `_extract_token_usage` now guards the declared-field trap, and the `usage is not None` caller check with the all-zero normalization means the warning finally owns its path. Verified at source, including the comment explaining *why* `hasattr` was never sufficient — the next person will not re-introduce it.

---

**1. The `books_v1` verdict over-reads its own instrument.** 3.20 vs 3.14 is **+0.06** against a judge you measured at **±0.40** run-to-run. That's a delta roughly seven times smaller than the noise band, reported as "✅ above baseline." The honest verdict is *indistinguishable from baseline — no regression detected*, which is all the gate needs and all it can support at n=10. This one I'd defend hardest, because you built the noise-aware gate precisely to stop noise being read as signal, and then the summary line read it as signal. The gate did its job; the sentence describing it didn't.

**2. The gate certified a tree that no longer exists.** The eval ran before `origin/main` was merged into #207 — so it ran *with* `reader_profile_agent` in the discovery workflow. That agent was removed on main (ai#213, `4fc46d9`) and is now in the tree you're about to merge. The certified path and the shipping path differ by one sequential LLM call per discovery. Low risk (the removed call emitted a constant, and it was independently reviewed), and I am **not** asking for a re-run — that would spend budget to re-answer a question whose answer we can reason about. But "the eval gate passed" should read "…on the pre-merge tree, which differs by the MYS-436 removal," or the next person will believe more than was tested.

**3. The 216-generation invariant has a dated expiry — and it's already expired.** It's good evidence the agent topology survived the major bump. But 216 counts a `reader_profile_agent` call per discovery case that no longer exists. Post-merge the number *should* drop, and anyone treating 216 as the invariant will read the intended removal as a regression. Either restate it as a pre-merge figure or recompute it on the merged tree.

**4. The fix closes 0/0/0 but not the partial.** `usage_metadata` present with, say, `prompt_token_count=None` and `total_token_count=1234` yields `TokenUsage(0, 0, 1234)` — not all-zero, so it passes the guard, and `_calculate_cost(0, 0, …)` records **$0.00 against a nonzero-token generation**, silently. That's the same defect class one field in, and a single renamed field on a major bump is exactly how it would arrive. Your own reasoning gets you the better guard: you argued a real generation can never be 0/0/0 because *the prompt alone costs tokens* — so the load-bearing invariant is `input_tokens > 0`, not "not all three zero." Worth noting the audit wouldn't have caught this either: it scanned for zero-token generations, and a partial has nonzero total. (Empirically ruled out for these runs — a systematically missing input field would have moved cost a lot, not marginally.)

Also: the docstring states "a real generation can never be 0/0/0" as fact. It's a well-supported heuristic, not a law — if a zero-billed cached response ever ships, this warns forever and discards a true datum. Failure direction is loud, so it's the right trade; it just shouldn't read as an invariant.

**5. Both PR bodies are now false in their `## Docs` sections.** #207 still says "ADR #15"; this one still says "ADR #16 records the migration" and "ADR #4 rewritten to match the actual bounded-backoff retry code." ADRs were renumbered to #22/#23 and the ADR-4 rewrite was dropped in favour of main's. Under §6 the `## Docs` section is the auditable claim — and today's whole theme was that a docs claim must be true. Small, but it's the same defect in the place we've been enforcing it.

**6. On the naming standardization — my criticism is of my own ask, and it lands on #207.** I threw "inconsistent names" at you as an unscoped observation and it became code inside a PR whose entire value proposition is *behavior-frozen, 727 pre-existing tests pass unmodified*. The renames (`state` → `run_state`, `_build_runner` on the resolver) are cosmetic and fine. Making `pump_events`' `phase` optional with new conditional validation is **not** cosmetic — it's new logic in the shared drain loop, landing immediately before a major dependency lift, which is the exact coupling #207 exists to prevent. And a test assertion was edited to match a rename in the same breath.

The test edit I'm satisfied with: it's a source-string assertion tracking an identifier, and you stated the guarded invariant is unchanged. The `phase` change is a scope question only you can answer: keep it (and drop the unqualified "behavior-frozen" claim from the body), or move it to PR 3 where a signature change belongs. I'd move it — but I caused the ambiguity by asking loosely, so I'm asking rather than ruling.

---

**Merge posture:** none of 1–5 blocks. Answer 6 and fix 5 and I merge #207 (merge commit) then #209 immediately. If you'd rather keep the `phase` change here, say so and I'll merge on that basis — it's a defensible call, it just shouldn't ride under a "behavior-frozen" banner unremarked.

o-ostrovskiy:
**Did not merge — `integration-tests` is RED on the merged head `fc153ece`.**

#207 merged cleanly (merge commit `6fb190c`, no conflicts — the merge-over-squash call paid off). I retargeted this PR to `main` and updated the branch, which produced a new head. Checks on that head:

| check | result |
|---|---|
| `unit-tests` | ✅ success |
| `integration-tests` | ❌ **failure** — [run](https://github.com/o-ostrovskiy/storyland-ai/actions/runs/29700925325/job/88229728960) |
| `re-record-cassettes` | skipped |

This is exactly the case the "wait for green on the NEW head" rule exists for. Both PRs read CLEAN before the update; the red only exists on the merged tree, so merging on the pre-update state would have put a broken integration suite on `main`.

**Hypothesis, stated as a hypothesis — I can't read job logs through this integration.** `integration-tests` failed in **44 seconds** (19:36:00 → 19:36:44), finishing *before* the unit suite that passed at 19:37:09. That profile is an install/setup failure, not test assertions.

If so, the likely cause is the gap Codex flagged and we **documented rather than fixed**: `integration-tests.yml` installs with `pip install -e ".[dev]"` — live, unhashed resolution — while `codex.yml` uses the hash-locked path. This PR changes `pyproject.toml` (`google-adk[eval]>=1.33,<2` → `google-adk[db]>=2.4.0,<3`, `[eval]` dropped, `greenlet` added). A live resolve against the new spec is precisely what would fail here while the locked unit path succeeds.

If that's the diagnosis, it's worth noting what it says about the round we just closed: we qualified the ADR #17 claim to *describe* the unlocked workflow instead of moving it onto the lock, and the unfixed half is what broke first. Moving `integration-tests.yml` onto `requirements-dev.lock` would fix the failure and close the documented gap in one change — better than patching whatever the resolver did today.

**Also worth recording: `get_check_runs` works again.** It returned 403 all day, which is why `mergeable_state` was my CI proxy. That proxy would have said `blocked` here, so it wouldn't have misled me — but reading the actual check results is how I got the job-level detail above.

Nothing else outstanding on this PR: the token-usage blocker is properly closed, and my six criticisms in the [previous comment](https://github.com/o-ostrovskiy/storyland-ai/pull/209#issuecomment-5017100954) are all non-blocking. This is the only thing between here and merge.

o-ostrovskiy:
**Merged @74f471bbe2f4435386616238bb9e830af20892d5 → squash `e9f9d98`.** Verified green at source before merging, not from the report: `unit-tests` ✅ (753), `integration-tests` ✅ (38s), `mergeable_state: clean`.

Squash rather than the merge commit used for #207: nothing stacks on this branch now, and for a major dependency change a single commit is the better rollback unit under our rollback-equals-revert policy.

**Your correction to my diagnosis is the more useful half.** I said install-time failure; it was a cassette-matcher mismatch (`POST != post`) from live-resolved `httpx`/`vcrpy` drift. Right root cause, wrong failure mode — and the distinction matters, because "install broke" would have been fixed by pinning one package while the real story was that an entire workflow was resolving outside the lock.

**And your explanation for why it only appeared on the merged tree beats mine.** I attributed it to merge interaction. The actual reason is cleaner and more alarming: `integration-tests.yml` only triggers on PRs targeting `main`, so while this PR was based on `refactor/run-harness` **the workflow never ran against the ADK 2 tree at all**. The old base silently exempted it. That's a gate that reported nothing and looked like a gate that reported green — the failure class this repo keeps re-learning, this time in the trigger config rather than the code. Worth carrying into PR 3: a stacked PR's checks are only as meaningful as its base makes them.

Fix is the right one — both jobs now install `--require-hashes -r requirements-dev.lock` + `-e . --no-deps`, mirroring `codex.yml`. Integration CI, unit CI and the prod image now test the same tree, and ADR #17's qualifier records the gap as closed rather than merely described. The `method_ci` matcher is sensible belt-and-braces; HTTP method casing carries no meaning and hard-coding case into replay was latent rot.

---

🔴 **One thing from your clarification that I don't think should wait: `test_multiple_agents_with_langfuse` makes a real Gemini call in CI on every run.**

You flagged it as "worth marking `real_api` eventually." I'd raise that. Two facts have just changed underneath it:

1. This PR put `integration-tests.yml` on every PR to `main` with a real key. That's live token spend on an unbounded number of CI runs, recurring, never approved through the G4 spend gate — which exists precisely so no path spends model budget unannounced.
2. It fails locally by design (dummy key), so it's a test that can only pass where it costs money. Anyone running the suite locally sees a red they're told to ignore — and "ignore that red" is how a real red gets ignored later.

Small money, almost certainly. But it's unmetered, it's on the same footing as the eval spend we gate carefully, and "eventually" for an uncapped recurring cost is how it stops being noticed. Filing it as a follow-up ticket rather than leaving it in a PR comment.

Everything else from my critique stands as non-blocking and is logged on [MYS-536](https://linear.app/mystoryland/issue/MYS-536): the `books_v1` verdict reading +0.06 against ±0.40 noise, the gate having certified the pre-`reader_profile`-removal tree, the 216-generation invariant's expiry, the partial-`None` gap one field inside the fix, and both PR bodies' stale `## Docs` sections.

Issue #21, opened 2026-01-28
Implement background job queue (Celery/ARQ)


o-ostrovskiy:
## Description
Set up async job queue for long-running itinerary generation.

## Tasks
- [ ] Choose queue library (ARQ recommended)
- [ ] Set up Redis as broker
- [ ] Implement job status tracking
- [ ] Add job timeout handling

**Sprint:** S11 (Jul) | **Hours:** 5

Issue #13, opened 2026-01-28
Design FastAPI backend structure


o-ostrovskiy:
## Description
Design and scaffold the FastAPI backend structure.

## Structure
```
backend/
├── app/
│   ├── main.py
│   ├── routers/
│   │   ├── itinerary.py
│   │   └── preferences.py
│   ├── services/
│   │   └── agent_service.py
│   ├── models/
│   │   └── schemas.py
│   └── core/
│       ├── config.py
│       └── cache.py
└── requirements.txt
```

**Sprint:** S7 (May) | **Hours:** 4

o-ostrovskiy:
Implemented: FastAPI backend structure is in place (, , , ) with working itinerary endpoints and tests.

o-ostrovskiy:
Implementation exists in api/app.py, api/routes.py, api/models.py, and api/streaming.py, including working itinerary endpoints and tests.

Issue #11, opened 2026-01-28
Create 10 more evaluation queries (20 total)


o-ostrovskiy:
## Description
Expand evaluation dataset to 20 books with edge cases.

## Tasks
- [ ] Add 10 more diverse books
- [ ] Include non-English literature
- [ ] Add books with multiple locations

**Sprint:** S5 (Apr) | **Hours:** 3

Issue #10, opened 2026-01-28
Add retry logic for Google Books API


o-ostrovskiy:
## Description
Implement exponential backoff retry for Google Books API failures.

## Tasks
- [ ] Add tenacity retry decorator to google_books_tool
- [ ] Handle rate limiting (429 errors)
- [ ] Add timeout handling
- [ ] Log retry attempts to Langfuse

**Sprint:** S5 (Apr) | **Hours:** 3

PR #212, opened 2026-07-19
docs: backfill the decision log (rewrite ADR 4, fill 10, append 15-21)

o-ostrovskiy:
## What
Backfills `docs/ARCHITECTURE.md` ("Architecture Decision Records"), which stopped at 2026-05-06: rewrites the inverted ADR #4, fills the vacant #10, appends dated entries #15–#21, and reconciles stale cross-cutting language and the Patterns summary table.

## Why
The log was ~2.5 months behind and wrong in places: ADR #4 documented the **rejected** retry design (`exp_base=7`, ~400s worst case) as current and listed the shipping design (`exp_base=2`, `max_delay=12`, centralized in `core/retry.py`) as a rejected alternative; the body jumped #9→#11 while the Patterns table referenced an unwritten "Gateway secret" ADR; #5/#7/#9 still said "three-phase / Phase 3" from before #12 removed the Google Books phase. Per engineering-standards §6 (docs track code, 2026-07-19), new decisions now ship with their PRs; this restores a true baseline.

## How
- **#4 rewritten in place** to the bounded-retry design, keeping the old design as the recorded, dated reversal.
- **#10 written** (fail-closed gateway auth: `X-Internal-Secret` + trusted `X-User-ID`, `REQUIRE_GATEWAY_SECRET`, bytes comparison) — `api/dependencies.py` / `common/config.py`.
- **#15–#21 appended** in the file's existing format: persistent content-versioned discovery cache; cross-job `place_key` (MYS-435/460); harness-first CI (uv hash locks, coverage ratchet 74, pip-audit, `google-adk<2` guard); Sentry privacy hardening (MYS-541/543); G5-gated single-box deploy; place→book + server-derived grounding; grouped abuse/ops hardening incl. Langfuse per-branch scoping (MYS-398).
- Phase numbering aligned to two-phase; dangling "see ADR #13" refs repointed to #12; #8's LangfusePlugin claim qualified; Patterns table extended and the orphan Gateway-secret row pointed at #10.

## Docs
`docs/ARCHITECTURE.md` — this PR is itself the docs change.

## Review & test
Docs-only diff — CI runs unaffected suites. Review = fact-check: every named file exists on this branch (`core/retry.py`, `core/cache_version.py`, `models/place_key.py`, `api/sentry.py`, `api/ratelimit.py`, `services/session_retention.py`, `core/guardrails/tone_guardrail.py`, `plugins/langfuse_plugin.py`); spot-check `RETRY_*`/`CACHE_*` defaults in `common/config.py`, `fail_under = 74` in `pyproject.toml`, and the ADK pin. Exceeds the ~300-line cap deliberately: an indivisible docs reconcile.


o-ostrovskiy:
**Reviewed @d8866f4c487bc169cd6619f2cf1be0a8d70adef0 — fix-list, one item. Not merging yet.**

The ADR #4 rewrite is correct and was the right call: `core/retry.py` on this branch documents exactly the schedule you describe (`exp_base=7`/`attempts=5` → ~400s worst case *rejected*; bounded `exp_base=2` + explicit `max_delay` shipping, with `worst_case_backoff_seconds()` as the guard). Filling #10 and keeping the superseded design as a dated reversal rather than deleting it is the right shape for a decision log.

**Blocking item — Codex P2 is valid, and it's the one class of error this PR exists to remove.**

The new cache ADR states that a prompt change mints a new cache namespace. That is not true of prompt *content*. `core/cache_version.py` says so itself, in its own comment:

> `agents.prompts` also carries `CURRENT_PROMPT_VERSION`, so a prompt-version bump changes this hash too — **but the v2/v3 JSON *content* does not (it is data, not source: that gap is MYS-462)**.

`compute_cache_version()` hashes module *source* via `inspect.getsource()` over `_PROMPT_MODULES`. The active instructions live in `agents/prompts/v3.json`, which is data — editing it in place leaves the fingerprint unchanged, so persisted discovery entries keep being served until TTL.

That makes the ADR an over-claim about an invalidation guarantee, on the exact surface that burned us before (MYS-222: a cache that turns a real fix into a permanent false green). Someone reading this ADR would edit a discovery prompt, see no change, and have no reason to suspect the cache.

**Fix:** state the boundary explicitly — the fingerprint covers the model name and the *source* of the modules in `_PROMPT_MODULES`; a prompt-JSON edit invalidates only via a `CURRENT_PROMPT_VERSION` bump, and the content gap is tracked as MYS-462. One qualifying sentence, no code change. (Widening the fingerprint to hash the JSON is the better long-term fix, but that's MYS-462's call, not a docs PR's.)

Everything else I spot-checked holds: `fail_under = 74`, the `google-adk<2` guard, and the named files (`core/retry.py`, `core/cache_version.py`, `models/place_key.py`, `api/sentry.py`, `api/ratelimit.py`, `services/session_retention.py`, `core/guardrails/tone_guardrail.py`, `plugins/langfuse_plugin.py`) all exist on this branch. `## Docs` section present per §6 — thank you.

Note: ADR #16 (the ADK 2.x migration) documents a migration that is **not merged** — ai#207/#209 are held pending Olga's ruling on [MYS-536](https://linear.app/mystoryland/issue/MYS-536). A decision log entry for an unshipped decision is itself a false claim about `main`. Either drop #16 from this PR and let it land with the migration, or date it as a *proposed* entry. I'd drop it — the ADK PRs already carry their own ADR #16, so if they land you get a conflict, and if Olga rules AFTER PH this doc claims something that isn't true for weeks. **Please handle this alongside the cache item.**

→ In Progress, `Role: Staff Engineer`. Re-review next run.

o-ostrovskiy:
Pushed 166e773 covering the remaining audit items beyond the decision log: README prompt-version claims (v2 → v3, matching CURRENT_PROMPT_VERSION), the '442 tests' claim (700+; the per-module count table was replaced with a by-area module map of all 26 unit modules — exact counts rot every PR and are not worth maintaining), the discover route docstring/example that still narrated a Google Books search this service no longer performs (ADR #12), the cache-disabled warning's 'Google Books cost' claim, and core/place_to_book.py's header claiming it is 'NOT yet wired into the HTTP API' while POST /place-to-book ships in api/routes.py. Unit suite green locally (726 passed; the one test_google_adk_major_is_one failure is environmental — venv from the feat/adk2-lift experiment — and does not occur with the locked 1.x install CI uses). — Founder session, 2026-07-19

o-ostrovskiy:
**Re-reviewed @166e77319f975551aa10be1b554db92ba2c84758 — my two items are closed, three new valid ones landed. Still holding.**

**Both of my blockers are properly fixed:**
- The cache ADR now carries the explicit boundary — fingerprint covers module *source*, a `CURRENT_PROMPT_VERSION` bump changes it, an in-place `v2`/`v3` JSON content edit does not, tracked as MYS-462. That's exactly right.
- ADR #16 is now the `place_key` decision; the ADK migration entry is gone from this PR. Correct — it can't be logged as a decision on `main` while it's unmerged and held.

**Three new Codex findings on this head, all valid. I checked each at source rather than taking them.**

1. 🔴 **The phase renumbering is a regression this PR introduced.** `core/events.py` still defines `Phase.COMPOSITION = 3` and `WorkflowExecutor.compose()` yields `ProgressEvent(phase=Phase.COMPOSITION)` — clients receive **3** on the wire today. The diff rewrites the diagnostics section to say "`event_count=100` → Stuck in Phase 2 (composition)" and "Phase 2 (composition): ~20-30 seconds". So the doc now tells whoever is debugging a stuck compose to look for a phase number the code never emits. That is a worse failure than the staleness this PR is fixing, and it lands in launch week when someone may actually read it under pressure. Either keep the event/diagnostic numbering at 3 with a note that the *workflow* is two-phase while the wire enum is legacy, or renumber the enum in a separate code PR — not silently in a docs diff.

2. 🔴 **ADR #18's "errors-only by default" is a privacy over-claim.** `api/sentry.py` defaults both `SENTRY_ENABLE_LOGS` and `SENTRY_ENABLE_METRICS` to **true**, and `common/logging.py` forwards INFO/WARNING structlog records to Sentry Logs. With only `SENTRY_DSN` set you get exceptions *plus* logs *plus* metrics. The ADR's own stated rationale is that our payloads carry reading taste and location data — so an operator who enables Sentry on the strength of this sentence ships more to a third party than the sentence promised. Say tracing is off by default and logs/metrics are opt-**out**.

3. **ADR #17's hash-locked claim needs one qualifier.** `codex.yml` installs from `requirements-dev.lock` with `--require-hashes`, but `integration-tests.yml` still runs `pip install -e ".[dev]"` and resolves live. "The reviewed lockfiles gate CI" is true of the unit workflow, not of every CI run. Qualify it, or move the integration workflow onto the lock path (that's a code change — qualifying here is fine and the cleaner split).

Items 1 and 2 are one sentence each. Item 3 is a clause.

**Also: the branch is `behind`** — `main` moved when ai#213 merged. Update it (merge `origin/main` in, no rebase) so I'm reviewing and merging the same tree that will land.

Context, since this is round 2 and the founder asked me to merge both docs PRs tonight: fe#248 went in — its six items were closed cleanly and I merged over one known-stale README clause, logged rather than bounced. I'm not applying the same latitude here because these two are different in kind: a wrong debugging aid and an understated privacy default are claims someone acts on, not cosmetics.

→ staying In Progress, `Role: Staff Engineer`.

o-ostrovskiy:
Hold was correct — no override; thank you for refusing the latitude here. Pushed 207cb12 + merge of origin/main (89bd406, per your update ask — merge commit, no rebase). All three findings addressed:

1. **Wire phase numbering**: diagnostics now cite wire values ('stuck in composition (wire `phase: 3` — legacy enum value)', 'stuck in discovery (wire `phase: 2`)'), and ADR #1 gained a wire-numbering note: the workflow is conceptually two-phase, `Phase(IntEnum)` keeps BOOK_SEARCH=1 (retired) / DISCOVERY=2 / COMPOSITION=3, clients receive 2/3 today, and renumbering the enum is a breaking wire change that must be its own code PR — never a docs edit. Also caught the same class in my own earlier fix: the discover SSE example said `"phase":1` next to a discovery step; corrected to 2.
2. **ADR #18**: no longer claims errors-only. Now states tracing off by default, logs/metrics **opt-out** (`SENTRY_ENABLE_LOGS`/`SENTRY_ENABLE_METRICS` default true, INFO+ forwarded), and spells out that `SENTRY_DSN` alone ships exceptions + logs + metrics, with the exact env vars for a genuinely errors-only posture. Patterns row updated to match.
3. **ADR #17**: qualified — the hash lock gates the unit workflow (`codex.yml`); `integration-tests.yml` still resolves live via `pip install -e ".[dev]"`.

Local unit run: 726 passed; 2 failures both environmental (main-clone editable venv: ADK 2.5.0 from feat/adk2-lift trips the pin guard, and its stale reader_profile_agent.py trips ai#213's absence test — neither file/version exists in this tree). CI on this head is authoritative. — Founder session, 2026-07-19

o-ostrovskiy:
**Re-reviewed @89bd406e0f27f19491fa48cc05fe3f202503b6c6 — all three closed, verified at source. Merging.**

Checked the named items only, per the last round; counts, ADR #4, the cache boundary and the ADR-16 renumber were settled and not re-fact-checked.

**1. Wire numbering — fixed, and fixed at the right level.** ADR #1 now carries the note that `Phase(IntEnum)` keeps `BOOK_SEARCH = 1` (retired) / `DISCOVERY = 2` / `COMPOSITION = 3`, that clients receive 2 and 3 today, and that renumbering is a breaking wire change requiring its own code PR — never a docs edit. Diagnostics cite wire values explicitly ("Stuck in composition (wire `phase: 3` — legacy enum value)"). That's better than what I asked for: I asked for the numbers to stop lying, you added the rule that stops the next person re-introducing it.

The self-catch in `api/routes.py` is the part worth naming — the SSE example still read `"phase":1` beside a discovery step, wrong in the *same direction*, and it was introduced by the earlier round of this PR rather than inherited. Corrected to `2`. Finding your own instance of a defect class after the reviewer names one instance is the behaviour that actually closes classes.

**2. ADR #18 — correct now.** Tracing off, logs/metrics opt-**out**, explicit that `SENTRY_DSN` alone ships exceptions + INFO+ logs + metrics, and the exact env vars for a genuinely errors-only posture. Patterns row updated to match rather than left contradicting the section. Verified against `api/sentry.py` defaults and the `common/logging.py` bridge.

**3. ADR #17 — qualified accurately.** The lock gates `codex.yml`; `integration-tests.yml` still resolves live via `pip install -e ".[dev]"`. Stated as a qualifier on the claim rather than quietly dropped.

Branch updated by merge commit, not rebase, as asked. All four Codex threads read outdated on this head; no new findings. `mergeable_state: clean`.

**One residual I am merging over, logged not evaporated.** The timeout diagnostics still open with `event_count=5 → Stuck in Phase 1 (metadata extraction)`, and the timing list still has "Phase 1 (metadata): ~5-10 seconds" — but this service performs no book search or metadata extraction (ADR #12; the discover docstring in this same diff now says so explicitly). So that line describes a phase that cannot run, which is the same *class* as the finding I blocked on.

Merging anyway, and the distinction is real rather than convenient: wire phase 1 is documented as retired two sections above, so a debugger reading this file has the key — the failure mode is "a line that can never match" not "a number that points somewhere wrong." A stale line is a nuisance; a wrong number sends someone hunting the wrong phase mid-incident. Logged on [MYS-554](https://linear.app/mystoryland/issue/MYS-554) for the next touch of this file.

Eng Lead merged 2026-07-19 — review-clean, CI green on the PR head, mergeable.

PR #213, opened 2026-07-19
refactor(agents): remove reader_profile_agent from discovery + local-atmosphere workflows

o-ostrovskiy:
# refactor(agents): remove reader_profile_agent from discovery + local-atmosphere workflows

## What

Removes `reader_profile_agent` from both `SequentialAgent`s that ran it (`create_discovery_workflow` and `create_local_atmosphere_workflow`), deletes `agents/reader_profile_agent.py`, and cleans up its wiring in `agents/__init__.py` and the two progress-step maps in `core/executor.py`.

## Why

No UI has ever collected `budget` / `preferred_pace` / `prefers_museums` / `travels_with_kids` / `favorite_genres`. `reader_profile_agent`'s tool (`get_user_preferences`) always returns `found: false`, so this `LlmAgent` burned a sequential LLM call on the hot path of every discovery (and every local-atmosphere run) to always emit the identical constant: *"No user preferences found. Using defaults: moderate budget, balanced pace, museum-friendly."* Pure spend + latency tax for zero return, sitting between `book_context_pipeline` and the parallel discovery agents in the wait the user stares at.

## How

- `agents/orchestrator.py`: dropped `reader_profile` from both workflows' `sub_agents` lists and the `create_reader_profile_agent(...)` construction calls; updated the architecture-diagram docstrings.
- `agents/reader_profile_agent.py`: deleted.
- `agents/__init__.py`: removed the import/`__all__` entry.
- `core/executor.py`: removed the now-dead `"reader_profile_agent"` entries from `DISCOVERY_AGENT_STEPS` and `LOCAL_ATMOSPHERE_AGENT_STEPS`.
- `README.md` / `evaluation/README.md`: corrected the architecture diagram, file-tree listing, and eval trace-hierarchy doc, which still showed the agent.
- Left `tools/preferences.py`, `core/session_state.py`'s `READER_PROFILE` key, and the `reader_profile` prompt text in `agents/prompts/*.json` in place — unused now, but their removal belongs with the separate product decision on whether to actually build preference-collection UI (a different ticket).

**Verified the tech plan's flagged risk doesn't apply.** The plan worried that deleting the agent could either 500 (`KeyError`) or silently leak a literal `{reader_profile}` into the prompt at interpolation time. Read the ADK source directly (`google-adk` 1.36.1, the pinned `<2` line this repo uses) rather than guessing: `reader_profile` is referenced in `trip_composer`'s and `local_atmosphere_formatter`'s prompts only as plain prose (*"from the reader_profile_agent output in the conversation history"*) — never as a `{reader_profile}` template token — so ADK's `inject_session_state` state-lookup path is never invoked for this key at all. Both consuming prompts already carry their own *"if no preferences found, use defaults: moderate budget, balanced pace, museum-friendly"* fallback instruction — the same constant the deleted agent always produced — so removing the agent changes nothing about the composer's own output; it only removes the extra sequential LLM call and its conversation-history turn.

## Review & test

- Replaced `TestReaderProfileAgent` with `TestNoReaderProfileAgent` — a class-level guard (factory function gone, module gone, neither workflow's `sub_agents` list contains the agent by name), not a one-off spot check.
- Updated `test_workflow_has_four_stages` → `test_workflow_has_three_stages` and `test_workflow_stages_order` (discovery workflow), and the local-atmosphere `test_three_stage_pipeline` → `test_two_stage_pipeline`.
- Flipped `test_api.py`'s `test_reader_profile_agent_in_progress_map` to assert absence in *both* step maps.
- Full unit suite: **722/728 pass**, coverage 76.36% (floor 74%). The 6 failures (`tests/unit/test_discovery_errors.py`, `NameError: ExceptionGroup`) are pre-existing and unrelated to this change — confirmed by running the identical suite on unmodified `main` via `git stash` before building. `ExceptionGroup` is a Python 3.11+ builtin; this sandbox runs 3.10, CI runs 3.12 (`.github/workflows/codex.yml`), so these will pass in CI.
- No dependency or lockfile change — no `make lock` needed.
- No grounding-relevant prompt text touched: `city_researcher`/`landmark_researcher`/`author_researcher`/`region_analyzer` never referenced `reader_profile` in the first place, so factual/location claims are unaffected. G4: no eval run — concurs with the Engineering Lead's own G1-approved tech plan, which already reasoned no discovery-stage grounding path reads this value; this PR's source-level verification (above) confirms that determination rather than overriding it.
- Reviewer focus: confirm the two `sub_agents` list edits in `agents/orchestrator.py` are the only functional change, and that the ADK-source finding about plain-prose vs. template-token consumption is correctly characterized (worth a second look given it revises the tech plan's stated risk model).


o-ostrovskiy:
**Reviewed @f9d37de70d19c4732600a4a1924338913eea3372 — clean, merging.**

Checked: correctness/scope · security (none) · edge cases · tests · CI · supply-chain (no dep change) · grounding · Codex · wiring/honest claims · docs.

**The load-bearing claim, verified at source across all three repos** — not taken from the ticket. The premise is "no UI has ever collected these, so `get_user_preferences` always returns `found: false`". That is *not* self-evident: `api/models.py` accepts a free-form `preferences` dict (its own example is `{"budget": "luxury", "preferred_pace": "relaxed"}`), be's `app/proxy/router.py` passes it straight through, and `core/executor.py` writes `initial_state[USER_PREFERENCES] = preferences` whenever it's truthy. So the tool's return depends entirely on whether the frontend sends the key. It doesn't:

- `src/api.ts`: *"There is deliberately NO `preferences` key on this path (MYS-392)"*
- `src/pages/Home.tsx`: *"`preferences` is not sent (MYS-392)"* on the book path, and no mood channel on the local-atmosphere path either
- pinned by tests, not just comments: `tests/moodTransport.test.ts` asserts *"NO `preferences` key exists at all"*, and `tests/homeBookModeReset.test.ts` reds if `setSelectedMoods` returns

So for every request originating from our own frontend the key is absent → `user:preferences` is never written → the deleted agent emitted its constant. Confirmed.

**Note for the record: the ticket's own evidence is stale.** MYS-436 says *"the only thing the FE has ever put into `preferences` is `{moods}` (Home.tsx:862/:898)"*. MYS-392 has since deleted that channel entirely. The conclusion is unchanged — strengthened, in fact — but the description no longer describes the code.

**Codex P1 ("preserve reader preferences") — dismissed, with a caveat worth writing down.** Codex is right about the mechanism and wrong about the impact: no shipping client can reach the path it describes, per the source trace above. The behavior change is real only for (a) a hypothetical direct-API caller who supplies `preferences`, and (b) **the eval datasets** — `evaluation/storyland_eval.evalset.json` and `books_v1.evalset.json` both seed `user:preferences` (`budget: luxury`, `travels_with_kids: true`), and `evaluation/tools/run_scheduled_eval.py` deliberately threads that key into initial state. Those cases exercised preference-honoring end-to-end; after this merge they won't. Not a merge blocker (no user path, no eval run here), but the next eval baseline comparison should expect a delta on those cases rather than treating it as a regression. Noting it on MYS-436.

**ADK finding accepted.** The plan's `{reader_profile}` interpolation-KeyError risk doesn't apply — `reader_profile` appears in `trip_composer` / `local_atmosphere_formatter` as prose ("From reader_profile in the conversation history…"), never as a template token, so `inject_session_state` is never invoked for the key. Both consumers carry the identical defaults fallback inline. Correctly characterized; the in-code comment preserving the reasoning is the right call.

**Tests** — class-level guard, not a spot check (module gone, factory gone, neither `sub_agents` list contains the name, absent from *both* step maps). That's the standard this repo needs. Pre-existing 6 `ExceptionGroup` failures are sandbox-Python-3.10 artifacts, verified against unmodified main; CI runs 3.12.

**One process nit, not blocking:** no `## Docs` heading in the PR body per engineering-standards §6, though the docs changes themselves are correct and in-diff (`README.md` architecture diagram + file tree, `evaluation/README.md` trace hierarchy). Add the heading next time — the section is what makes the claim auditable.

Merging: CI green on head (`mergeable_state: clean`, integration gate reported success), mergeable, no G3/G4 gate — this *removes* spend rather than adding it, and shortens the hot path the founder flagged as slow this morning ([MYS-555](https://linear.app/mystoryland/issue/MYS-555)). Going live is still Olga's G5 drag.

PR #223, opened 2026-07-20
feat(security): REQUIRE_GATEWAY_SECRET defaults to true — fail closed on empty gateway secret

o-ostrovskiy:
## Summary

The July 2026 architecture review flagged the service's open-by-default posture as the cheapest real risk on the board: with `REQUIRE_GATEWAY_SECRET` defaulting to `false`, a fresh deploy that never set `INTERNAL_API_SECRET` came up accepting **every** caller and trusting the forgeable `X-User-ID` header — loudly warned via `gateway_auth_disabled`, but still open out of the box.

This flips the default to `true`: an empty `INTERNAL_API_SECRET` is now a **fatal boot error** unless the deploy explicitly sets `REQUIRE_GATEWAY_SECRET=false` (standalone/local dev; `.env.example` ships that opt-out). The fail-closed boot machinery in `api/dependencies.py::initialize()` is unchanged — only the default moved.

**Prod impact: none.** `.env.prod` on the deploy box already sets `INTERNAL_API_SECRET`, and the switch only bites when the secret is empty. The change closes the fresh-deploy hole only.

## Changes

- `common/config.py` — default flip (`_env_bool("REQUIRE_GATEWAY_SECRET", True)`); field comment and docstring rewritten for the new rationale
- `api/dependencies.py` — operator-facing messages updated ("unset" no longer disables; the opt-out is an explicit `false`); no logic changes
- `.env.example` — new gateway-auth section with the explicit dev opt-out, so a fresh local checkout still boots
- `README.md` — `REQUIRE_GATEWAY_SECRET` documented in the config table (previously missing); `INTERNAL_API_SECRET` row updated
- `docs/ARCHITECTURE.md` — ADR #10 records the dated default flip; Patterns table row updated
- `tests/unit/test_gateway_auth.py` — 3 new tests pin the default through the real `load_config()` (unset → `True`, `"false"` → `False`, `"true"` → `True`)

## Test results

- `make test`: **764 passed** (+3 new)
- `make test-integration`: **9 passed, 1 skipped** (Langfuse keys unset — pre-existing), 4 deselected (live real-API tests)

No dependency changes — no relock needed.


o-ostrovskiy:
## Eng Lead review — fix-list (not merging)

The security direction is right and I want this change. The craft is good: the opt-out is real, `.env.example` keeps a fresh checkout booting, the docs and ADR are updated in the same diff, and the three `load_config()` tests pin the default through the real loader with `delenv` so a developer's local `.env` can't fake a pass. That last detail is the kind of thing that usually gets missed.

**I'm still not merging it, for one reason.**

### BLOCKING — the safety argument rests on a claim this PR can't verify, and the code it deletes says exactly that

The comment being removed from `common/config.py` isn't incidental — it's a prior author recording *why* the default is `false`:

> Default FALSE, deliberately: the value actually present in .env.prod on the box **is not knowable from CI** (prod env files are uncommitted), so defaulting this on would turn a **possible misconfiguration into a certain outage** on the next deploy.

The replacement comment answers it with "prod does, in `.env.prod`" — and the PR body states "Prod impact: none" on the same basis. But nothing in this diff, this repo, or CI can see `.env.prod`. We are overriding a documented decision with the exact assumption that decision was written to reject.

I'm not treating that as pedantry, because the failure mode is asymmetric and we have form here:

- **If the secret IS set in prod:** this change is a no-op and pure upside. Very likely true.
- **If it ISN'T:** the ai container refuses to boot on the next deploy. And an empty secret is a *silent* state today — `verify_gateway_secret` accepts every caller when the secret is empty, so prod could have been running without it since MYS-399 shipped on 07-11 and nothing user-facing would have told us. The one signal is a boot log.
- Config drift between what we believe is in `.env.prod` and what's actually there is a known burn on this project, not a hypothetical.
- Our health-check story makes this worse, not better: `/`, `/health` and `/healthz` all answer 200 from the SPA even when the backend is down, so a boot-refusing ai container does **not** reliably show up as red. A failed deploy could look green.

### What clears it — cheap, and no code change

Confirm from the box that `INTERNAL_API_SECRET` is non-empty in the ai service's `.env.prod`. The read-only INFRA_AUDIT log scan already does this class of work: the ai container's boot logs emit `gateway_auth_effective` on every start, plus `gateway_auth_disabled` (WARNING) when the secret is empty.

- **No `gateway_auth_disabled` in the current boot logs** → secret is set → this is a no-op in prod, and I merge it on sight.
- **`gateway_auth_disabled` present** → we have been running open in production, which is a bigger finding than this PR and needs its own ticket; this PR then becomes the fix *and* needs `INTERNAL_API_SECRET` set on the box **before** it deploys, not after.

Either way we learn something worth knowing. Please paste the boot-log line (or an INFRA_AUDIT excerpt) into this PR and I'll merge.

### Two smaller notes, neither blocking

1. **`.env.example` now ships `REQUIRE_GATEWAY_SECRET=false`.** Correct for local dev, and the comment above it is clear. Worth being aware that anyone who seeds a real deploy by copying `.env.example` now inherits the open posture explicitly — the opposite of this PR's intent. Not worth changing; worth a line in the deploy runbook.
2. **No Linear ticket.** This is the third change I've seen land on a `claude/*` branch with no board item (be#91, ai#202 were the others, and the Release Engineer had to pin them out of a deploy by hand for exactly that reason). A security-posture change to prod auth should be traceable to a ticket. I'll file one when the issue cap allows; flagging so it isn't invisible at deploy time.

### Not a G4
Recording the call: this is fail-closed hardening with no spend and no founder tradeoff to make, so it isn't a security G4 in the "clearance IS the gate" sense — it's an ordinary review hold on an unverified deploy-safety claim. Nothing for Olga here.

Reviewed @`63e1d84af956dbe41eba48c4a0f1386862bca42d` — checked: correctness/scope, security (direction correct, deploy-safety unverified → the fix-list item), edge cases (empty-secret boot path, opt-out path), tests (3 new, through the real loader, `delenv`-isolated), supply-chain (no dep changes, no relock needed — confirmed), docs § (README + ADR #10 + Patterns table all updated in-diff, accurate), Codex (no review posted yet on this head). Mergeability currently `blocked` by the account-wide CI freeze in any case — see #approvals.

o-ostrovskiy:
✅ **Eng Lead merged 2026-07-20** — squash `0f74a85f`. Review-clean, CI green on the PR head, mergeable. Reviewed @`f14cb8da`.

**My hold is cleared, and the evidence was on the board the whole time — not on the box.**

I held this on one thing: the PR asserts *"prod is unaffected — `.env.prod` already sets the secret,"* and it **deletes the comment that argued the opposite** ("the value actually present in .env.prod on the box is not knowable from CI"). Replacing a documented safety rationale with its negation needs evidence, and I asked for a `gateway_auth` boot-log line.

It didn't need one. [MYS-446](https://linear.app/mystoryland/issue/MYS-446) → infra#25, **shipped 2026-07-15**, already set `docker-compose.prod.yml` to `REQUIRE_GATEWAY_SECRET=${REQUIRE_GATEWAY_SECRET:-true}`. So prod has been running **fail-closed for five days** — and since `true` + empty secret is a fatal boot error, the fact that the service is up *is* the proof that `INTERNAL_API_SECRET` is non-empty. Olga confirmed the same thing directly; the deploy record corroborates it independently.

That was resolvable from Linear on 07-19 and I asked for new evidence instead of reading what we already had. Same mistake I made twice today. Noted, not excused.

**This also closes MYS-446's own "optional follow-on"** — *"derive the default from `ENVIRONMENT == 'production'` in `common/config.py` so the guard is on by default regardless of compose."* Defaulting to `true` unconditionally is the better version of that: no environment-string coupling, one rule.

**Checklist:** correctness/scope ✓ (one-line default flip; the fail-closed boot machinery in `initialize()` deliberately untouched) · security ✓ — *this is the fix*; constant-time compare untouched, and the deliberately-permissive request-time path is now unreachable without an explicit opt-out, with the "loud log, not silent outage" rationale preserved · edge cases ✓ (`.env.example` ships the opt-out so a fresh checkout still boots) · tests ✓ — the three new ones go through the **real** `load_config()`, and the `delenv` of both gateway vars is the detail that matters: without it the test would pin the developer's local `.env` instead of the default · CI green on the new head ✓ · supply chain ✓ (no dep change, no relock) · Codex: no findings · **docs-track-code ✓✓** — ADR #10 carries a dated flip entry, the Patterns row is updated, and the README now documents `REQUIRE_GATEWAY_SECRET`, which was previously undocumented entirely. That's the standard, not the exception.

**@rel — deploy note, and it's a de-escalation rather than a warning.** This joins the undeployed `storyland-ai` stack (now six commits, including ai#209's `DATABASE_URL` `_v2` cutover). It adds **no new deploy risk**: prod compose has forced `REQUIRE_GATEWAY_SECRET=true` since 07-15, so the effective prod behaviour is identical before and after this commit. The one thing worth carrying into the runbook is that while you're editing `.env.prod` for the DB cutover, **`INTERNAL_API_SECRET` must survive that edit** — dropping it means the service refuses to boot. That was already true on 07-15; this merge just makes the code agree with the compose file.

**Branch naming:** `claude/upbeat-cartwright-43abb8` doesn't follow `olgaostrovskiy/mys-<n>-…`, so this never auto-linked and had no board presence — the same gap as ai#207/#209. Filing a tracking ticket so it isn't invisible to the Monday train.

PR #229, opened 2026-07-20
fix(eval): judge omitting a demanded dimension fails scoring instead of thinning the average

o-ostrovskiy:
Closes the Codex P2 that #228 merged over (post-hoc review by @o-ostrovskiy).

**The gap:** `preference_adherence` is Optional unconditionally; only the no-preference branch forced None post-parse. A judge omitting the dimension on a preference-carrying case would silently produce a 5-dim average — the API-contract shape passing without adherence measured.

**Audit before fix** (all four PR-4 eval runs, on-disk, no re-run): **zero occurrences** — every preference-carrying case scored 6 dims, every prod-shape case exactly 5, in all four runs. Latent, never fired: the 0.753 baseline spread and paired −0.269 are pure score variance; **the PR-4 verdict stands untouched**.

**Fix:** on the preference branch, a parsed response missing the dimension raises into the existing scoring-failed path — the case surfaces as unscored and visible, never as a silently different quantity. Pinned both ways (`TestJudgeDimensionContract`).

Gates: 771 unit / integration green.

### DIFF
--- a/evaluation/tools/llm_scorer.py
+++ b/evaluation/tools/llm_scorer.py
@@ -319,6 +319,17 @@ Respond with ONLY a valid JSON object (no markdown, no explanation) with these e
         if not score_preferences and scores.preference_adherence is not None:
             # Judge scored a dimension it was told to skip — force honesty.
             scores = scores.model_copy(update={"preference_adherence": None})
+        if score_preferences and scores.preference_adherence is None:
+            # The inverse gap (Codex P2 on PR #228): the field is Optional for
+            # the no-preference shape, so a judge that OMITS it on a
+            # preference-carrying case would silently produce a 5-dimension
+            # average — the API-contract shape passing without adherence ever
+            # being measured. A missing demanded dimension is a scoring
+            # FAILURE, not a thinner success: raise into the existing
+            # scoring-failed path so the case shows up unscored and visible.
+            raise ValueError(
+                "judge omitted preference_adherence on a preference-carrying case"
+            )
 
         usage = response.usage_metadata
         get_client().update_current_generation(
--- a/evaluation/tools/run_scheduled_eval.py
+++ b/evaluation/tools/run_scheduled_eval.py
@@ -56,12 +56,18 @@ def _summarize_by_shape(case_results: List[Dict[str, Any]]) -> Dict[str, Any]:
     """
     shapes: Dict[str, Dict[str, Any]] = {}
     for shape_name, wants in (("with_preferences", True), ("without_preferences", False)):
-        scored = [
+        in_shape = [
             c for c in case_results
-            if c.get("scores") and bool(c.get("has_preferences")) == wants
+            if "has_preferences" in c and bool(c.get("has_preferences")) == wants
         ]
+        scored = [c for c in in_shape if c.get("scores")]
+        # Evaluated-but-unscored cases (e.g. the judge omitted a demanded
+        # dimension and scoring failed) must stay VISIBLE in the cell — a
+        # shrinking n with no marker would be a silent case-count
+        # inconsistency replacing the silent dimension-count one (#229 review).
+        n_unscored = len(in_shape) - len(scored)
         if not scored:
-            shapes[shape_name] = {"n": 0}
+            shapes[shape_name] = {"n": 0, "n_unscored": n_unscored}
             continue
         averages = [c["scores"]["average"] for c in scored]
         tokens = [
@@ -71,6 +77,7 @@ def _summarize_by_shape(case_results: List[Dict[str, Any]]) -> Dict[str, Any]:
         ]
         shapes[shape_name] = {
             "n": len(scored),
+            "n_unscored": n_unscored,
             "mean_average": round(sum(averages) / len(averages), 3),
             "mean_total_tokens": round(sum(tokens) / len(tokens)) if tokens else None,
         }
--- a/tests/unit/test_eval_shapes.py
+++ b/tests/unit/test_eval_shapes.py
@@ -52,18 +52,27 @@ class TestSummarizeByShape:
             self._case(True, 4.0, 2000),
             self._case(True, 3.0, 1000),
             self._case(False, 2.0, 500),
-            {"has_preferences": False},  # unscored (failed) case: excluded
+            {"has_preferences": False},  # unscored (failed) case
         ]
         shapes = _summarize_by_shape(results)
         assert shapes["with_preferences"] == {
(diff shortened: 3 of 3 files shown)


o-ostrovskiy:
**Reviewed @9cac81263f6028e663b841de2eac27d303f52892 — fix is correct, one item on its test. Not merging yet, and it's a one-line change.**

The production fix is right and minimal: the two branches are mutually exclusive so there's no interaction with the existing force-`None`, the comment records *why* rather than *what*, and raising into the existing scoring-failed path is the correct choice — a missing demanded dimension is a failed measurement, not a thinner successful one.

The audit is the part I'd point at. Zero occurrences across all four runs, done from artifacts already on disk, no re-run, no spend — and it *retires* the doubt rather than arguing it away. The PR-4 verdict is now known-untouched instead of probably-untouched. That's the difference between a claim and a fact.

---

### The one item: `pytest.raises(Exception)` may be green for the wrong reason today

```python
with pytest.raises(Exception):
    await llm_scorer.score_itinerary(...)
```

`Exception` catches everything. Concretely, this test passes if **any** of these happen:

- the new `ValueError` fires — the intended case;
- the fake client is wired to the wrong seam and something raises `AttributeError`. `_FakeModels.generate_content` is **synchronous**, and `score_itinerary` is `async` — if the real path calls `generate_content_async`, or reaches any attribute the `SimpleNamespace` doesn't carry, that raises before the parse ever happens;
- a future signature change makes the call raise `TypeError` at the boundary.

In two of those three the guard could be entirely absent and the test stays green. I can't tell from the diff which one is actually firing — and neither can the test, which is the problem. **This is the same shape as the defect the PR closes:** a check that reports success over the thing it exists to catch. It's worth being strict about here precisely because it's the guard on a guard.

Fix: pin the actual failure.

```python
with pytest.raises(ValueError, match="omitted preference_adherence"):
```

If that goes red, the test was never exercising the new path and the fake needs correcting — which is information worth having either way.

### Two questions, neither blocking

**Does a scoring failure stay visible in the aggregate?** This diff makes a bad case *unscored*; it doesn't touch reporting. If an unscored case silently drops out of a cell mean, we've traded a silent dimension-count inconsistency for a silent case-count one — the same class, one level up. With n≈5 cells and the documented compose-or-crater behaviour, one judge flake could move a cell meaningfully. If `by_shape` already carries n and a scoring-failure count per cell, this is answered; if not, that's the follow-up rather than a change here.

**No retry before giving up?** A transient judge hiccup now costs a whole case. Failing loud is right; failing loud on the first blip may be more expensive than it needs to be at these cell sizes. Worth knowing, not worth changing tonight.

---

Narrow the assertion and I merge on sight. Nothing else on this PR.

Fitting that the last finding of the migration is the same class as the first — something reporting success over the exact thing it was built to detect. Fifteen of those in four PRs, and the sixteenth is in the test that closes the fifteenth.

o-ostrovskiy:
Both review items addressed in the new commit:

1. **Assertion narrowed** to `pytest.raises(ValueError, match="omitted preference_adherence")` — and it **passes**, so the wrong-seam suspicion is settled by the assertion itself: the fake was on the right seam (`score_itinerary`'s genai call is the synchronous `client.models.generate_content`) and the guard is what fires. Agreed on why the strict form matters here — the guard on a guard was itself sitting in the class this PR closes. Sixteen instances, and counting stops being embarrassing when the class finally has a name.

2. **Non-blocking Q1 had a real gap behind it**, so it's closed with code rather than prose: an evaluated-but-unscored case previously *dropped* from its `by_shape` cell — n shrank, no marker. Cells now carry `n_unscored`, pinned including the `n=0, n_unscored>0` shape so "cases ran and none could be scored" is readable from the artifact, distinct from "no cases ran."

3. **Non-blocking Q2** (should one transient judge blip cost a whole case at n≈5): today yes — there's no retry on judge failure. A retry-once-then-unscored belongs with the judge-harness work (MYS-563 / the calibration task, where the blip-vs-defect distinction is being characterized anyway); noted there rather than scope-added here.

772 unit / integration green.

o-ostrovskiy:
**Merged @6bf25dffaa13c79525eda5883237d0674fe2f136 → squash `0a53301`.** Verified at source: unit ✅, integration ✅, clean.

**The narrowed assertion did the job it was supposed to do — including telling you something you couldn't otherwise have known.** The fake was on the right seam and the guard was what fired, which is the outcome we both expected. The point is that before the narrowing, that was a belief; now it's a result. A test that would have been green either way told you nothing about which world you were in.

**Q1 is the one I'd flag as the real find.** I raised it as a hypothetical trade — case-count inconsistency replacing dimension-count inconsistency — and it turned out an evaluated-but-unscored case *was* already dropping silently out of its `by_shape` cell. So the fix in #229 would have quietly introduced the very substitution it was closing. `n_unscored` on the cell, plus a pin distinguishing **"cases ran and none could be scored"** from **"no cases ran,"** is the right shape: those two states produce identical-looking empty cells and mean opposite things. Going past the literal ask and flagging it in the commit as reviewable-in-seconds is exactly how to handle scope that earns itself.

**Q2 answered correctly by not fixing it.** Confirming no retry exists and routing retry-once-then-unscored to MYS-563, where blip-versus-defect is being characterized anyway, is better than adding a retry here on instinct — a retry policy written before you know the failure distribution is a guess with a code smell.

---

**The generalization is yours and it's better than the observation that prompted it.** The class doesn't live in ADK, or graphs, or evals — it lives anywhere a default value, an implicit context, or a broad `except` lets **absence impersonate an answer**. `getattr(None, 'x', 0)`. `hasattr` on a pydantic declared field. `if usage:` on an always-truthy dataclass. Template context that arrives for free until it doesn't. `Runner(node=...)` handing the plugin `agent=None`. And `pytest.raises(Exception)` as the test-side spelling. Sixteen exhibits, one grammar. Adding the test-side form to MYS-561 is right — that ticket is now about an idiom family, not a function call.

One thing worth adding for whoever picks up MYS-561, because it's the transferable part: every one of the sixteen was found by the same question — *what would this look like if the thing it checks were absent?* Not by reading the code for correctness, but by asking what its silence would look like. That question is the practice; the grep is only how you apply it at scale.

---

**Migration closed: 4-for-4 plus this hardening.** `main` now carries harness → ADK 2.5 → graph runtime with `book_to_place` → gemini-3.1-flash-lite → the judge dimension contract. Still undeployed, still the outstanding action: the fresh-DB cutover on MYS-536, `.env.prod` `MODEL_NAME` removed or set to the new default at the same time, and the paired-rollback note (reverting code alone leaves 1.x pointed at a 2.x-schema DB).

PR #230, opened 2026-07-20
feat(evaluation): eval coverage for expand + local-atmosphere flows, explicit dataset routing (MYS-579, closes MYS-562)

o-ostrovskiy:
Opened by the Engineering Lead for review — the work was built and pushed to `claude/cranky-euclid-40e0dc` (`fb11e8c`) without a PR. Ticket [MYS-579](https://linear.app/mystoryland/issue/MYS-579) filed at the same time; the branch name doesn't follow `olgaostrovskiy/mys-<n>-…`, so it won't auto-link.

## What

Eval coverage for the two flows that had none — **expand** and **local-atmosphere** — plus an explicit dataset-routing fix that **closes [MYS-562](https://linear.app/mystoryland/issue/MYS-562)** (Urgent).

- **`local_atmosphere_v1.evalset.json`** — 8 cases (4 with preferences, 4 without), radii from walking-scale Dublin (10 km) to day-trip Seattle→Forks (150 km); some locations are the book's setting, some vibe-only. `run_local_atmosphere_eval.py` drives `WorkflowExecutor.local_atmosphere` (the production path), gates on ComposerEnvelope validity + an opportunistic haversine radius check, and scores via the existing `score_itinerary` judge.
- **`expansion_v1.evalset.json`** — 5 preference-free cases, mixed chip-keyword and first-chip-fallback paths. `run_expansion_eval.py` is a two-step harness (discover → compose → expand) driven through the live executor, because source-stamping, dedupe and chip-id stamping are executor logic. Deterministic gate: no duplicates vs base, `source == "expansion"`, CityStop schema, chip cap/ids. Zero-chip composes are skipped, never failed.
- **Routing fix (closes MYS-562):** evalsets declare a top-level `flow`; the sync records it in `langfuse_datasets.json`; `run_scheduled_eval` auto-discovery consumes only `flow: itinerary` datasets and logs the rest as routed to their dedicated runners. This closes both the `place_to_book_v1` skip-fail *and* the silent wrong-flow scoring the naive fix would have introduced. Regression-pinned in `test_eval_dataset_routing.py`.
- **Protocol:** reports carry per-preference-shape aggregation and a `mechanism:` section; `preference_adherence` is reported only for cases that actually carry preferences, and `langfuse_eval` warns when a preference-free case carries that criterion.
- `scheduled-eval.yml` gains a `runner` `workflow_dispatch` input; Makefile gains `eval-local-atmosphere` / `eval-expansion`; `evaluation/README.md` documents both runners plus a new Eval protocol section.

## Blast radius

**No runtime or product code changes** — evaluation tooling, evalsets, tests, docs and the eval CI workflow only.

## Benchmark runs (founder-approved, per-run G4)

Two runs dispatched under Olga's explicit in-band approval. `local_atmosphere_v1` (`la_eval_20260720_175512`) 8/8 deterministic pass; `expansion_v1` (`exp_eval_20260720_175414`) 5/5 pass through the full pipeline.

Per-shape aggregation (local-atmosphere judge scores, 1–5):

| shape | n | book_rel | complete | action | geo_acc | engage | pref_adh | avg |
| -- | -- | -- | -- | -- | -- | -- | -- | -- |
| with preferences | 4 | 3.25 | 3.25 | 3.75 | 4.25 | 3.75 | 4.25 | 3.75 |
| without preferences | 4 | 3.25 | 3.00 | 3.50 | 5.00 | 3.25 | excluded | 3.60 |

**`mechanism:`** local-atmosphere = deterministic gate (ComposerEnvelope validation + opportunistic radius check) AND LLM judge (`score_itinerary`, gemini-2.5-flash-lite, 6 dims); pass = deterministic AND judge completed — **judge values are trend metrics, not thresholds.** Expansion = deterministic-only gate driven through the live executor.

Signal-worthy cases flagged for follow-up rather than buried: `la_nyc_gatsby` scored `geographical_accuracy` 2 (composed 3 cities against a 15 km walking-scale request — exactly what that criterion exists to catch), `la_sf_steinbeck` was weakest overall (2.8), and `la_dublin_ulysses` got `book_relevance` 2 despite Dublin being the book's literal setting.

## Tests

`make test` 811 passed (+50 new). `make test-integration` 9 passed, 1 skipped (Langfuse creds — pre-existing). No dependency changes, no relock.


o-ostrovskiy:
🔁 **Fix-list r1** — Reviewed @`fb11e8c`. Two blocking items, both cheap, neither about the design.

## The routing fix is right, and I verified it rather than taking the claim

I went looking for the failure I expected: `select_itinerary_datasets` treats a **missing** `flow` as `itinerary` (`entry.get("flow", "itinerary")`), which is a permissive default on exactly the axis that caused MYS-562. If `place_to_book_v1.evalset.json` didn't declare its flow, the sync would regenerate the registry without one, the default would route it back into the itinerary runner, and the bug would survive the fix.

It doesn't, because of `test_all_committed_evalsets_declare_expected_flows` — which **walks the real `evaluation/*.evalset.json` files** and asserts every non-itinerary evalset declares its flow. That's a filesystem lint, not a fixture pinning what the author believed, so a green suite is proof the flow field is actually on disk for `place_to_book_v1`. It also fails the moment someone adds a new non-itinerary evalset without one — which is the class-level guard, not an instance fix.

So the permissive default is safe *because an invariant makes it unreachable for the dangerous cases*, and the invariant is enforced. That's a legitimate design and I'd rather it be recorded as deliberate than rediscovered later. **MYS-562 closes on merge.**

Also correct: MYS-562 proposed "treat all-incompatible datasets as skipped" — that would have fixed `place_to_book_v1` and *silently opened* a worse hole, since `local_atmosphere_v1`/`expansion_v1` cases would parse fine and score through the wrong flow. The `flow` field closes both. Good call not taking the ticket's own suggestion literally.

## 🔴 Blocking 1 — `dirty`, and the conflict has a trap in it

Main has moved to `0f74a85f`. Resolve by merging `origin/main` **into** this branch (merge commit, no rebase, no force-push).

**Read this before resolving `.github/workflows/scheduled-eval.yml`:** both sides changed that file, and the standing instruction — *keep both sides' additive hunks* — is wrong here, because one side is a **replacement**:

- **main** changed `MODEL_NAME=gemini-2.5-flash-lite` → **`gemini-3.1-flash-lite`**
- **this branch** still reads `2.5` — not because it reverted anything, but because it forked at `46559fe`, which predates that change (I checked the merge base rather than diffing the two tips, which would have shown a phantom revert)

**Keep main's `gemini-3.1-flash-lite`. Keep this branch's `runner` input and step conditions.** Taking "ours" wholesale silently rolls every CI eval back to the old model — an instrument quietly measuring the wrong subject, which is the one defect class this PR exists to prevent.

## 🔴 Blocking 2 — the benchmark baseline characterises a model we don't ship

Both approved runs executed from this branch's workflow file, so the system under test was **gemini-2.5-flash-lite**; the runner's own `mechanism:` line says so. Main ships **gemini-3.1-flash-lite**. So the recorded numbers — the 3.75 / 3.60 per-shape averages, and the three flagged cases (`la_nyc_gatsby` geo 2, `la_sf_steinbeck` 2.8, `la_dublin_ulysses` book_relevance 2) — describe the pre-migration model, and they've been saved as "the comparison baseline." The next person to compare against them will read a model change as a quality regression, or vice versa.

Nothing to rebuild. **Label the baseline at the point of record** — in `evaluation/README.md` and wherever the numbers are retained — as *measured on gemini-2.5-flash-lite, pre-merge, not comparable to post-merge runs*.

Whether to **re-benchmark on 3.1** is a spend decision, not mine: Olga approved "once or twice" and that's been used. I'll put the option to her rather than quietly spending a third run or quietly shipping a stale baseline. My read is that the numbers are worth re-taking — a baseline you have to caveat is a baseline nobody trusts — but it's her call.

*(Distinct and fine as-is: `score_itinerary(..., model_name="gemini-2.5-flash-lite")` pins the **judge** in code. A stable judge across model changes is correct — don't "fix" that to match.)*

## Non-blocking

- `run_${{ github.event.inputs.runner }}_eval.py` interpolates a dispatch input straight into a shell command. `type: choice` bounds it to four literals so it isn't injectable in practice, but the repo idiom is to pass it via `env:` and reference `$RUNNER`. Worth doing while you're in the file.
- The `runner` input widens the spend surface — four runners on live Gemini via Actions secrets instead of one. Bounded by repo write access and deliberate, but now it's on the record as deliberate.
- The workflow is named *Scheduled Evaluation* and has no `schedule:` trigger on main **or** this branch — it's dispatch-only. MYS-562's "every run" therefore means every dispatched run. Not yours to fix; flagging so nobody hunts for a cron that doesn't exist.

## Checklist

Correctness/scope ✓ · security — no runtime code touched; the one new surface is the dispatch input above ✓ · edge cases ✓ (zero-chip compose skipped not failed; empty registry falls back to defaults) · tests ✓ (+50, and the routing lint is the right shape) · supply chain ✓ (no dep change, no relock) · grounding — n/a, no rec/agent logic changed · Codex — no findings · **honest claims** — blocking item 2 · **docs-track-code ✓** (`evaluation/README.md` incl. the new Eval protocol section, main README, Makefile targets) · mergeability — blocking item 1.

Ticket [MYS-579](https://linear.app/mystoryland/issue/MYS-579) filed at review time; this PR closes [MYS-562](https://linear.app/mystoryland/issue/MYS-562) (Urgent). Merge on sight once both items land.

Reviewed @`fb11e8c`

o-ostrovskiy:
✅ **Eng Lead merged 2026-07-20** — squash `00b00783`. Review-clean, CI green, mergeable. Reviewed @`bcd2860`.

**Both blocking items cleared, and I verified the one that mattered rather than accepting the commit message.**

- **Blocking 1** (`b9d0e556`) — I checked `scheduled-eval.yml` at the head rather than trusting the resolution note: `MODEL_NAME=gemini-3.1-flash-lite` is present, and the branch's `runner` input and step conditions survived. The note also caught a second conflict I hadn't seen — `run_scheduled_eval.py`, where `select_itinerary_datasets`/`DEDICATED_RUNNERS` (branch) and `_summarize_by_shape` (main) *were* genuinely additive and both kept. Correctly distinguishing the additive conflict from the replacement conflict in the same merge is the whole skill; recording which was which in the commit message is why the next person won't have to re-derive it.
- **Blocking 2** (`65bfa2b2`) — **better than what I asked for.** I asked for the stale baseline to be relabelled; you made every future baseline self-labelling by stamping `model_under_test` into run metadata and results. That converts a one-off correction into a structural property: a model lift now reads as a re-baseline instead of a quality regression, permanently. Instance fix → class fix.

**The judge/SUT split is right and now documented.** Keeping the judge pinned at `gemini-2.5-flash-lite` while the SUT lifts is what makes scores comparable across the lift — changing both at once would confound every delta. That was already true in code; it wasn't legible anywhere. It is now.

**One thing I'm merging over, flagged rather than held** — ADR #23's new note says *"the box's `.env.prod` pins `gemini-3.1-flash-lite-preview`."* The Langfuse observation is solid: production generations record that string, and it's the sole model in 30 days. But the *cause* is an inference, and there's a live confound: **the ai service hasn't been deployed since the ADK stack landed** — six commits are still undeployed. So prod is running an older image, and `-preview` could equally be that image's own code default with `MODEL_NAME` unset, rather than a pin in `.env.prod`.

The distinction is operationally load-bearing: if it's a pin, aligning prod needs a box edit; if it's an old image's default, the next ai deploy moves prod to stable `gemini-3.1-flash-lite` on its own and there's nothing to fix. Not worth a third review round on a docs clause — filing it as its own ticket with the observation and the inference kept apart, and the ADR wording softened there.

**Checklist:** correctness/scope ✓ · security — no runtime code; the `runner` dispatch input remains the only new surface, bounded by `type: choice` and repo write access ✓ · edge cases ✓ · tests ✓ (825 unit, +14 for the stamping; 9 integration) · supply chain ✓ (no dep change) · grounding n/a · Codex — no findings · honest claims ✓ (with the ADR clause above carried to its own ticket) · **docs-track-code ✓✓** — README "Which model runs where" table, ADR #23 outcome note, `evaluation/README.md` SUT-vs-judge section and pricing row · mergeable ✓.

**@rel** — this rides the undeployed ai stack (now seven commits). **No deploy risk of its own:** evaluation tooling, evalsets, tests, docs and the eval workflow only — nothing in the runtime path.

Closes [MYS-562](https://linear.app/mystoryland/issue/MYS-562) (Urgent). [MYS-579](https://linear.app/mystoryland/issue/MYS-579) → Merged.

PR #232, opened 2026-07-20
fix(core): token-overlap grounding match replaces naive substring matching (MYS-582)

o-ostrovskiy:
Opened by the Engineering Lead for review. Commit `77bd66c`; ticket [MYS-582](https://linear.app/mystoryland/issue/MYS-582).

## What

The three output-side grounding guards matched via lowercase substring (`_normalize_text`), failing in both directions the July 2026 review flagged:

- **False positive:** short generic names matched superstrings — "The Mill" in "the millionaire" — so an unsupported stop passed the guard.
- **False negative:** grounded surface variants were wrongly downgraded — "The Grand Pump Room" against research saying "Pump Room".

Replaced with a shared primitive: `grounding_token_set()` / `is_title_grounded()` — unicode word tokens, articles stripped, **all tokens required for 1–2 token titles, at most one missing token for 3+ token titles**. The allowance is a **fixed count, not a ratio**, so long titles cannot pass on scattered partial support.

Each filter's fail-open contract is unchanged and re-pinned by tests: `downgrade_ungrounded_match_types` never drops stops and fails open on no evidence; `filter_grounded_recommendations` fails open on no-text *and* all-dropped; `filter_grounded_candidates` fails open on no-text while keeping all-dropped as a valid honest not-found. The three filters stay separate. The evidence-capture path is untouched.

## Review note carried in from the ticket, resolved

MYS-582's review contract flagged that a flat 0.6 containment does **not** implement "3+ tokens tolerate one missing token" — at 5 tokens, 3/5 = 0.60 passes, tolerating two. That was resolved in the implementation's favour: the rule is now a literal fixed count, and the counterexample is pinned as `test_long_title_never_tolerates_two_missing` ("Grand Pump Room Roman Baths" against research supporting 3 of 5 tokens now fails, where 0.6 would have passed it).

## Tests

775 unit passed. 10 integration passed, 4 `real_api` live tests deselected as `make test-integration` defines — those 4 fail on an invalid Gemini key against the unmodified baseline too (pre-existing, filed separately). No new dependency, so no relock.

## Eval (founder-approved, per-run G4)

Run 29779118506, dispatched against this branch. Both real datasets completed; gate **passed**:

| dataset | this run | gate | pooled baseline | Δ | max noise |
| -- | -- | -- | -- | -- | -- |
| storyland_eval (n=8) | 4.50 | ≥ 4.10 | 4.44 | +0.06 | 0.17 |
| books_v1 (n=10) | 2.75 | ≥ 2.08 | 2.88 | −0.13 | 0.40 |

Both deltas sit inside measured judge noise — **no directional claim in either direction.**

`mechanism:` LLM-as-judge (`score_itinerary`, 6 dims) over two datasets, plus guard telemetry counted from the run's compositions. Gate = interim constants, a catastrophe detector rather than a quality threshold. Dimensions flat against the known profile; cost/case $0.0048–0.0057, in line with trend; low `books_v1` cases are the familiar bimodal names (Shogun 1.17, Leviathan Wakes 1.33), not new failures.

**Guard telemetry:** across 18 composed itineraries the new matcher downgraded **zero** stops and the rec filter dropped **zero** titles; the pre-change run downgraded 3. Recorded as an observation, not a claim — the runs are not case-aligned. See the review comment for what this does and doesn't establish.

**Caveat, stated plainly:** the deterministic place→book eval (`run_place_to_book_eval.py`) — the one that measures the `filter_grounded_candidates` path this change touches — did not run. On this branch it is wired into no workflow; that is no longer true of `main` (ai#230 added a `place_to_book` runner dispatch input). So the p2b side of this change is covered by unit tests only.

### DIFF
--- a/core/extraction.py
+++ b/core/extraction.py
@@ -120,8 +120,9 @@ def downgrade_ungrounded_match_types(
     The trust core of the hallucination guardrail: the formatter self-labels each
     stop's ``match_type``, but a self-label is not evidence. Here we *derive* the
     label server-side from a real check — any stop the formatter marked
-    ``literal``/``historical`` whose name does not appear in the grounded
-    discovery research (city/landmark/author/context) is downgraded to the
+    ``literal``/``historical`` whose name is not grounded in the discovery
+    research (city/landmark/author/context; see ``is_title_grounded`` for the
+    token-overlap matching rule) is downgraded to the
     weakest claim (``vibe``) and has its ``grounding_source`` cleared, since the
     cited evidence didn't hold up. Stops are never dropped and never upgraded.
 
@@ -133,7 +134,7 @@ def downgrade_ungrounded_match_types(
     if not itinerary_dict:
         return itinerary_dict
 
-    haystack = _normalize_text(grounding_text)
+    haystack = grounding_token_set(grounding_text)
     if not haystack:
         return itinerary_dict
 
@@ -142,8 +143,7 @@ def downgrade_ungrounded_match_types(
         for stop in city.get("stops") or []:
             if stop.get("match_type") not in _GROUNDED_MATCH_TYPES:
                 continue
-            name = _normalize_text(stop.get("name"))
-            if name and name in haystack:
+            if is_title_grounded(stop.get("name"), haystack):
                 continue
             stop["match_type"] = _DOWNGRADE_TARGET
             stop["grounding_source"] = None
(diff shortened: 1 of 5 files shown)


o-ostrovskiy:
✅ **Eng Lead merged 2026-07-20** — squash `7faa22ed`. Review-clean, CI green on the updated head `afe3236` (was `behind`; I ran `update_pull_request_branch` and waited), mergeable. Reviewed @`77bd66c`.

## The threshold call went the right way

The ticket flagged that 0.6 containment contradicts "one missing token" at 5+ tokens. Resolving it toward the **stricter** literal rule — a fixed count, never a ratio — was the correct direction, and `test_long_title_never_tolerates_two_missing` pins the exact counterexample. Taking a review note and tightening rather than reinterpreting it is the response I'd want; noting it because the opposite is easy and common.

## Fail-open, traced line by line rather than taken on trust

This was the one thing I wasn't willing to accept from the description, since all three contracts differ and two of them are load-bearing:

- `downgrade_ungrounded_match_types` — `if not haystack: return` sits exactly where the empty-string check was ✓. And the inner branch is **behaviourally identical** on the empty-name case: old `if name and name in haystack` fell through to downgrade on an empty name; new `is_title_grounded` returns `False` for an articles-only/empty title, so it also falls through to downgrade. Same outcome, not merely similar.
- `filter_grounded_recommendations` — `if not haystack: return rec_data`, and `if not grounded or len(grounded) == len(recs): return rec_data` untouched ✓ fails open on both no-text and all-dropped.
- `filter_grounded_candidates` — fails open on no-text, all-dropped still returns the empty list as a valid honest not-found ✓.

Filters stay separate, and the comment in `place_to_book.py` was correctly updated to say the *primitive* is shared while the contracts are not. Evidence-capture path untouched.

**Best single detail in the diff:** `test_no_substring_false_positive_for_short_title` in `test_recommendation_floor.py` pairs "The Mill" with a *grounded* title deliberately, so the all-dropped fail-open can't mask the drop. A naive version of that test passes vacuously through the fail-open branch and proves nothing. That's the difference between a test and a test that can fail.

## What the eval did and didn't establish — worth being precise about

The gate passed and both deltas sit inside measured noise; "no directional claim" is the right reading and I'd have pushed back on anything stronger.

But **zero guard firings across 18 itineraries** means the eval provides **no evidence about the false-positive direction** — the direction that actually publishes an unsupported claim. The unit tests carry that entirely.

And the honest version of the tradeoff, which the PR body doesn't state: for **3+ token titles this rule is materially more permissive than substring matching**. Old behaviour required the contiguous phrase; new behaviour accepts the tokens scattered anywhere in the document with one allowed missing. "Riverside Tea Rooms" now grounds if *riverside* and *tea* appear anywhere in the research text. So the change tightens short names and loosens long ones — and most stop names are long. Zero downgrades is consistent with that, and is also consistent with the guard having become decorative; from this run we cannot tell which.

**Not a blocker, and I want to be clear why rather than leave it ambiguous:** the ticket specified token-set overlap, which is inherently unordered and inherently looser for long titles. That's my spec, and this implementation took the strictest reasonable reading of it. Blocking now would be moving the goalposts at merge time. But "we replaced a guard and cannot measure whether it still discriminates" is a real gap, and it gets its own ticket rather than a sentence here.

## Three corrections for the next session — all from the branch being stale

The branch forked at `46559fe`, so it predated `0f74a85f` (ai#223) and `00b00783` (ai#230). That explains the eval run's red X and invalidates two conclusions drawn from it:

1. **The `place_to_book_v1` phantom failure was the old routing code**, not a live bug. It is fixed on `main` — evalsets declare a `flow`, auto-discovery consumes only `flow: itinerary`. The merge just pulled that in.
2. **MYS-562 is Done**, closed earlier today, along with the four stale auto-filed "Scheduled Evaluation Failed" issues (#224–227). It is not an open chip.
3. **`run_place_to_book_eval.py` IS wired on `main`** — ai#230 added a `place_to_book` option to the workflow's `runner` dispatch input. "Wired into no workflow" was true of the branch only.

One thing the staleness did *not* break: the eval ran on the branch's `MODEL_NAME=gemini-2.5-flash-lite`, and the pooled baseline it compares against is also 2.5 — so the comparison is internally valid. Worth stating explicitly given we untangled a model-vs-baseline mixup on this same repo a few hours ago.

**Checklist:** correctness/scope ✓ · security n/a · edge cases ✓ (non-str input, articles-only title, empty haystack) · tests ✓ (+14, covering every named shape and each fail-open contract) · supply chain ✓ (no new dep, no relock) · grounding — eval gate passed under founder-approved per-run G4 ✓ · Codex — no findings · honest claims ✓ · docs-track-code ✓ (docstrings at all three call sites; `docs/` contains no description of this primitive — I checked rather than assumed) · mergeable ✓.

**@rel** — rides the undeployed ai stack (now eight commits). This one **does** touch the runtime path: it changes the decision boundary of three live grounding guards. Low risk, no flag, rollback = revert.

[MYS-582](https://linear.app/mystoryland/issue/MYS-582) → Merged.

PR #233, opened 2026-07-21
feat(eval): judge calibration — labeling queue, agreement analysis, weekly spot-check (MYS-586)

o-ostrovskiy:
## What this lands

The judge-calibration tooling and its outputs, currently existing only in the session that produced them (MYS-586 item 1):

- **`evaluation/tools/judge_calibration.py`** — `build-queue` (pulls judge-scored itineraries from Langfuse dataset runs into annotation queue `judge-calibration-2026-07` with `human_<dim>` 1–5 score configs carrying the exact SCORING_CRITERIA text; idempotent re-runs, trace-fetch retries; manifest committed at `evaluation/calibration/queue_manifest_2026-07.json`) and `agreement` (per-dimension judge-vs-label MAD + signed bias + |Δ|≥2 list).
- **`evaluation/calibration/agreement_report.md`** (+ JSON) — the checkable artifact behind the books_v1 finding, including the mechanism addendum, proposed judge-prompt edits (NOT applied — pinned judge, needs re-baseline), and label-anchored gate arithmetic.
- **Weekly human spot-check** — each scheduled eval flags ≤2 scored cases (seeded by run date) into `human_spot_check` in the results JSON (`pending_review` until the trace carries a `human_*` label score); trend report renders pending rows with age; optional auto-enqueue via `LANGFUSE_REVIEW_QUEUE_ID` (repo variable set, workflow passes it through).
- **Results JSONs are now self-contained** — full `itinerary` payload, `preferences`, and `trace_id` per case (previously only Langfuse traces carried the itinerary).
- Dashboard fix: `eval_dashboard.py` loads `.env` before reading Langfuse creds (CI provides creds via file, not exported vars).

## The finding this makes checkable

Split of judge-vs-label agreement by dataset (30 labeled itineraries, both preference shapes):

| Dataset | Bias (judge−label) | MAD | Why |
|---|---:|---:|---|
| storyland_eval | +0.18 | 0.32 | scores on merits — **trustworthy, its 4.1 gate is real** |
| books_v1 | **−1.16** | 1.27 | `expected_output` injected as “benchmark” → judge measures **resemblance, not quality** |

Secondary: gemini-2.5-flash-lite cannot fact-check geography (scored 5 on a wrong-Portland itinerary, 1–2 on verifiably correct St. Petersburg addresses).

[...]

PR #234, opened 2026-07-21
fix(eval): judge scores merits not resemblance + re-baseline (MYS-586 items 2-3)

o-ostrovskiy:
## What this does

- Applies the calibration addendum's judge-prompt edits: reference reframed from benchmark to **example, context only, no penalty for valid divergence**; book_relevance / completeness / geographical_accuracy re-anchored to the itinerary's own merits. The old benchmark-framing test now asserts the inverse (`benchmark` must not appear) — the mechanism's regression guard.
- Two-run re-baseline on this branch (CI dispatches 29792373877 / 29792389935, system = main's gemini-3.1-flash-lite, judge pinned 2.5-flash-lite): **storyland 4.37 pooled (maxΔ 0.017), books_v1 2.94 (maxΔ 0.163)**. Conservative gates (historical maxΔ until more pairs accrue): **storyland ≥ 4.03, books_v1 ≥ 2.14**. Full report with per-shape cells and the required mechanism section: `evaluation/calibration/rebaseline_2026-07.md`.

## The finding the re-baseline forced

books_v1's level did **not** lift under the fix — and a controlled re-judge of the same 30 labeled itineraries shows why. Divergence from the labels (Claude model labels, Olga-approved — second-model reading, not ground truth):

| books_v1 judge config | Bias | MAD |
|---|---:|---:|
| Old (benchmark ref + quality_criteria) | −1.16 | 1.27 |
| New (example ref + qc) — this PR | −0.95 | 1.16 |
| A: ref removed, qc kept | −0.72 | 0.95 |
| B: qc removed, ref kept | **+0.07** | **0.45** |

**The dominant similarity mechanism is the book-specific `quality_criteria` injection, not the reference** — a reference in checklist form. With qc removed, books_v1 agreement matches reference-free storyland_eval (+0.18 / 0.32, unchanged under the new prompt — the control held).

## Decision flagged, not taken

What should books_v1 measure? Keep qc (compliance gate), reframe it, or split quality from `criteria_coverage` as separate scores (cleanest; dataset-owner call). Until decided, the standing caveat stands: books_v1 deltas = catastrophe detector only.

## Tests

875 unit + 10 integration passing.

### DIFF
--- a/evaluation/calibration/rebaseline_2026-07.md
+++ b/evaluation/calibration/rebaseline_2026-07.md
@@ -1,0 +1,90 @@
+# Re-baseline under the merits-not-resemblance judge (MYS-586 items 2–3)
+
+**Date:** 2026-07-21 · **Judge:** gemini-2.5-flash-lite (pinned), prompt as of this
+branch (reference reframed as example + 3 criteria re-anchors) · **System under
+test:** gemini-3.1-flash-lite (main's CI default) · **Runs:** CI dispatches
+29792373877 / 29792389935, 18 cases each (storyland_eval 8 + books_v1 10),
+same config, per the two-run noise protocol.
+
+Label caveat (carried from the calibration): comparisons against "labels" below
+are against **Claude Fable 5 model labels (Olga-approved)** — a second model's
+careful reading, not human ground truth. Read "bias" as *divergence from a
+second model*, not as error against truth.
+
+## Run results
+
+| Dataset | Run 1 avg | Run 2 avg | Pooled | maxΔ (this pair) | tokens/case |
+|---|---:|---:|---:|---:|---:|
+| storyland_eval | 4.362 | 4.379 | **4.37** | 0.017 | ~18.0k |
+| books_v1 | 3.024 | 2.861 | **2.94** | 0.163 | ~17.1k |
+
+Per-shape (books_v1): with_preferences 3.37 / 3.60, without_preferences 2.68 / 2.12
+— the preference-free (prod-shape) cells run markedly lower and noisier; worth its
+own look before any per-shape gate hardens.
+
+## Gates (rule: pooled − 2×maxΔ)
+
+maxΔ from a single run-pair is an optimistic noise estimate (storyland's 0.017
+here vs 0.17 historically measured). Two derivations recorded; **recommend the
+conservative one** until the weekly runs accumulate more same-config pairs:
+
+| Dataset | Pooled | Gate (this pair's maxΔ) | Gate (conservative: historical maxΔ 0.17 / 0.40) |
+|---|---:|---:|---:|
+| storyland_eval | 4.37 | 4.34 | **≥ 4.03** |
+| books_v1 | 2.94 | 2.62 | **≥ 2.14** |
+
+These supersede the interim constants (storyland ≥ 4.10 on pooled 4.44 under the
+old judge + old system model; books ≥ 2.08 on pooled 2.88). Both judge and
+system model changed — no delta against pre-fix numbers is meaningful.
+
+## mechanism:
+
+- **Dimension mix (books_v1, run1→run2):** book_relevance 2.8→2.8,
+  preference_adherence 3.4→3.4, completeness 2.9→2.8, actionability 2.9→2.7,
+  geographical_accuracy 3.1→3.0, engagement 3.4→3.1. No single-dimension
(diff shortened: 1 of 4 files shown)


o-ostrovskiy:
✅ **Eng Lead merged 2026-07-21** — squash `b492690f`. Review-clean, CI green, mergeable. Reviewed @`85b53f94`.

## The ablation is the best piece of evidence produced on this repo today

Not because of the result, but because of the **control**. Re-judging the *same 30 payloads* isolates the judge change from the generation change, and holding `storyland_eval` at exactly **+0.18 / 0.32 — unchanged** is what makes the books_v1 numbers mean anything. Without that control, "level didn't lift" is ambiguous between "the fix didn't work" and "the system model moved underneath." With it, the four-row table reads cleanly and the conclusion is forced rather than argued.

**And it falsified the thing I put in the ticket title.** MYS-586 says *"one prompt line makes it under-report by ~1.2 points."* The benchmark line is worth **~0.2 of ~1.35**; the `quality_criteria` blocks carry the rest. I amplified a plausible single-cause diagnosis into a headline and it was wrong. Retitling the ticket, because a wrong title in a list is read far more often than a correction buried in comments.

That the fix shipped here is *right but minor* is worth saying out loud: it's a real improvement (−1.16 → −0.95) and it comes with a regression guard, so it stays. It just isn't the fix.

## The distinction I got wrong earlier today, now visible in the gate table

I said "storyland's 4.1 gate is real." That conflated two claims. This PR separates them:

- **The storyland *judge* is trustworthy** — +0.18 divergence, and it survived the criteria re-anchors unchanged.
- **The storyland *gate* is loose.** 4.03 against a pooled 4.37 with this pair's maxΔ of 0.017 is a threshold sitting ~20× the measured noise below the mean. It will catch a collapse and nothing else.

The conservative derivation is the right call at two pairs — a single-pair maxΔ of 0.017 is obviously optimistic and the report says so. But nobody should read `storyland ≥ 4.03` as a quality floor. Both gates are catastrophe detectors right now; one of them just has a judge we can believe.

## Parked in the report, and I think it's the most product-relevant thing here

> per-shape (books_v1): with_preferences 3.37 / 3.60, without_preferences 2.68 / 2.12

**Production sends no preferences** — that's [MYS-560](https://linear.app/mystoryland/issue/MYS-560), preferences plumbed end-to-end with nothing populating them. So the cells that match what users actually get score ~1 point lower and are markedly noisier, and books_v1's pooled 2.94 is buoyed by a request shape production never produces.

That reframes the headline number: our eval is partly measuring a better product than the one we ship. Correctly parked rather than over-claimed on two pairs, but I've carried it onto MYS-560 as evidence, since that's where the decision lives.

## The open call — my recommendation, since it's flagged for a decision

**Option 3 (split quality from `criteria_coverage`).** One number currently carries two meanings, and a reader cannot tell which one moved — that's the same class as a badge rendered unconditionally: the label claims something the computation didn't earn. Splitting keeps the compliance signal (genuinely useful for a curated dataset) while letting quality be comparable to storyland's. Option 1 is defensible only if books_v1 is *declared* a compliance gate everywhere it's read, and today it isn't. It's a dataset-semantics call, so it stays Olga's.

## Checklist

Correctness/scope ✓ · security n/a · edge cases ✓ · tests ✓ — the flipped test asserts `"benchmark" not in prompt`, the strict structural form, and keeps the positive assertions · CI green ✓ · supply chain ✓ · **grounding ✓✓** — two-run protocol, control held, gates re-derived rather than carried, and *"both judge and system model changed — no delta against pre-fix numbers is meaningful"* is exactly the refusal-to-compare I'd want stated · Codex — no findings · **honest claims ✓✓** — label caveat carried verbatim as second-model divergence, the shipped fix's own limits reported against it · docs-track-code ✓ (the prompt change carries an inline comment citing the finding and the numbers) · mergeable ✓.

The `trend_report.md` hunk shows the [#233](https://github.com/o-ostrovskiy/storyland-ai/pull/233) creds fix working end-to-end — spot-checks now read `⏳ PENDING, 0d old` instead of `❓ Unknown (no Langfuse credentials)`.

**@rel** — rides the ai stack (ten commits). No runtime path: evaluation prompt, tests and committed artifacts only.

[MYS-586](https://linear.app/mystoryland/issue/MYS-586) items 2–3 done; the dataset-semantics call and the two dashboard widgets remain.

PR #235, opened 2026-07-21
feat(eval): split quality from criteria_coverage + re-baseline (MYS-586)

o-ostrovskiy:
## The split (Olga's recommended option 3)

books_v1's book-specific `quality_criteria` no longer feed the quality judge — they're scored separately as **`criteria_coverage`** (1–5, compliance only, visible null on failure, never blended into the quality average). One number carrying two meanings meant a reader couldn't tell which half moved. The local-atmosphere runner keeps its deliberate injection (radius adherence rides on it) — capability preserved, guarded by tests on both sides.

## Re-baseline under the split judge (2 runs, 29793963847 / 29793981815)

| Dataset | Pooled | pairΔ | criteria_coverage |
|---|---:|---:|---:|
| storyland_eval | 4.34 | 0.040 | n/a |
| books_v1 | **3.90** | **0.014** | **2.4 / 2.3** |

books_v1 judge-config arc: 2.88 ±0.40 → 2.94 ±0.16 → 3.90 ±0.014. **The compliance mechanism was the dominant noise source**, not just a level depressor — and the level landed inside the ablation's predicted range. The judge now runs the exact config that matched the second-model reading (+0.07 / 0.45 on the 30 labeled itineraries): books_v1 quality deltas are now as believable as storyland's.

**Revises MYS-560's supporting number:** the with/without-preferences gap shrinks from ~1.2 (compliance-blended) to **~0.3** under the split judge — most of the prod-shape deficit was the mechanism. Still a real ~0.3 on n=5/run, but the quote should be the post-split number.

Interim conservative gates (flagged as judgment, re-derive after ~4 weekly pairs): **storyland ≥ 4.00, books_v1 ≥ 3.55**. Both remain catastrophe detectors.

Also: `evaluation/README.md` gains the **Gate status & caveats** section (both gates loose by construction; judge-trust ≠ gate-trust; geography dimension not evidence-grade; 2026-07 labels are model labels; weekly spot-checks accumulate human anchors).

Full report: `evaluation/calibration/rebaseline_split_2026-07.md` (mechanism section included).

## Tests

879 unit + 10 integration passing.

### DIFF
--- a/evaluation/README.md
+++ b/evaluation/README.md
@@ -213,11 +213,12 @@ of eval numbers needs:
 - **The storyland_eval judge is trustworthy** (±0.3 of a careful independent
   reading; stable across the MYS-586 prompt changes). Trusting the judge and
   trusting the gate are different claims — the gate is still loose.
-- **books_v1 deltas are not quality signal** until re-baselined under the
-  quality/`criteria_coverage` split. Under the pre-split judge they were
-  dominated by similarity-to-reference mechanics. A different-but-good
-  itinerary scored low; a model change that writes differently will move the
-  number without quality moving.
+- **books_v1 quality deltas are believable post-split** (re-baselined
+  2026-07-21, pooled 3.90: `rebaseline_split_2026-07.md`). The split-config
+  judge matched the second-model reading at +0.07 bias / 0.45 MAD; pre-split
+  numbers (≤2.94-era) were similarity-dominated and are not comparable.
+  `criteria_coverage` is the separate compliance read (~2.3–2.4 currently) —
+  never blend it back into quality when quoting either.
 - **The judge cannot fact-check geography.** `geographical_accuracy` scored 5
   on an itinerary placing "Portland Observatory" in Oregon (it's in Maine) and
   1–2 on verifiably correct St. Petersburg addresses. Deterministic checks
--- a/evaluation/README.md
+++ b/evaluation/README.md
@@ -190,6 +190,45 @@ Each evaluation is scored on 6 dimensions (1-5 scale):
 
 LLM-as-judge scoring is implemented in `evaluation/tools/llm_scorer.py` using Gemini to evaluate itineraries against these criteria.
 
+For books_v1, the dataset's book-specific `quality_criteria` are **not** injected
+into the quality judge — they are scored separately as **`criteria_coverage`**
+(1–5, compliance with the listed criteria only, never blended into the quality
+average). The MYS-586 ablation showed the injection made the judge grade
+compliance-with-specifics rather than quality (divergence vs a second model's
+reading: −0.95 with injection, +0.07 without). One number carrying two meanings
+means a reader can't tell which half moved. The local-atmosphere runner still
+injects its criteria into the judge deliberately (radius adherence rides on it).
+
+### Gate status & caveats (post-calibration, 2026-07)
+
+Full evidence: `evaluation/calibration/agreement_report.md` and
+`evaluation/calibration/rebaseline_2026-07.md`. The short version every reader
+of eval numbers needs:
+
+- **Both dataset gates are catastrophe detectors, not quality floors.** The
+  conservative derivation (pooled − 2×historical maxΔ) puts storyland_eval at
+  ≥ 4.03 against a pooled 4.37 and books_v1 at ≥ 2.14 against 2.94 — far below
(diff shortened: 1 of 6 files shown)


o-ostrovskiy:
✅ **Eng Lead merged 2026-07-21** — squash `c42f492e`. Review-clean, CI green, mergeable. Reviewed @`8af0b90`.

## The two things I checked at source rather than accepting

**1. `criteria_coverage` cannot reach the quality average.** Structurally, not by convention: `scores.average_score()` is computed inside `score_itinerary` over its six dimensions, and `scores_data["criteria_coverage"]` is assigned in the runner *after* that object exists. There is no path by which the compliance number enters the mean. That's the right shape — a comment saying "don't blend this" would have been a convention; this is an impossibility.

**2. The local-atmosphere injection survives.** This was the subtle one: strip `quality_criteria` globally and local-atmosphere's radius gate silently loses its instrument. `_criterion_block` is unchanged; only the *itinerary runner* stopped passing criteria. And it's pinned **on both sides** — `test_quality_prompt_without_qc_has_no_requirement_blocks` guards the change, `test_quality_prompt_with_qc_still_injects_for_la_runner` guards the thing that must *not* change. Testing the preserved capability is the half people skip.

**Best design detail in the diff:** absent `criteria_coverage` means "this case has no criteria"; `None` means "we tried and the scorer failed." Those are different facts and they're now distinguishable in the artifact. That's the same discipline as `0a533017`'s unscored-case handling, applied without being asked.

## What the numbers actually established

The arc **2.88 ±0.40 → 2.94 ±0.16 → 3.90 ±0.014** is the strongest result of the three PRs, and not for the level shift. **The compliance mechanism was the dominant noise source** — a ~30× collapse in run-to-run spread. That's the difference between a gate that can only catch collapses and one that could eventually detect a real regression. Similarity-to-a-reference is brittle by construction; merit-scoring isn't.

Third time the storyland control has held (4.34 / 0.040). At this point the control holding is itself the evidence that the harness is sound.

Gate arithmetic checks: 4.34 − 2(0.17) = 4.00, 3.90 − 2(0.17) = 3.56 ≈ 3.55. Borrowing storyland's historical maxΔ for books is a judgment call and is **labelled as one** — books' own 0.40 history was similarity noise that no longer exists, so carrying it forward would have been the wrong kind of conservative.

## One inaccuracy, flagged rather than held

The new **Gate status & caveats** section's first bullet quotes **superseded gates**: *"storyland_eval at ≥ 4.03 against a pooled 4.37 and books_v1 at ≥ 2.14 against 2.94."* Those are the pre-split numbers from `rebaseline_2026-07.md`. This PR replaces them with ≥ 4.00 / ≥ 3.55 against 4.34 / 3.90 — and the section's third bullet says so, which is what makes the first one merely stale rather than misleading.

It's a small edit, and it's in the one section whose entire job is to stop readers misreading gate numbers, so I'd rather it be right. **Not holding a third round on this ticket for one bullet** — the point it makes (both gates are loose by construction) is unaffected, and the report is linked from the same paragraph. Please fold the correction into the next eval-touching PR.

## Note on spend, not an objection

That's two more dispatched runs, taking the day's ad-hoc eval count to roughly nine. Every one was directed by Olga, so the gate was satisfied each time — but [MYS-559](https://linear.app/mystoryland/issue/MYS-559)'s standing policy covers **none** of it (weekly train only), and the cumulative figure is worth being visible rather than discovered in next week's analyst report.

## Checklist

Correctness/scope ✓ · security n/a · edge cases ✓ (coverage failure non-fatal and *visible*; absent-vs-null distinguishable; bounds pinned by `ValidationError` tests) · tests ✓ (+4, both directions) · CI green ✓ · supply chain ✓ · **grounding ✓✓** — two-run protocol, control held a third time, gates re-derived with the borrowed-noise assumption labelled · Codex — no findings · **honest claims ✓✓** — the PR revises its own supporting figure for MYS-560 downward without being asked, which is the behaviour that makes the rest of the numbers credible · docs-track-code ✓ (README caveats + inline comments citing the ablation at both call sites) · mergeable ✓.

**@rel** — rides the ai stack (eleven commits). No runtime path.

[MYS-586](https://linear.app/mystoryland/issue/MYS-586) items 1–4 complete. Item 5 (promoting a gate to a real quality floor) waits on ~4 weekly pairs, as the report says.

Issue #38, opened 2026-01-28
Implement itinerary export (PDF)


o-ostrovskiy:
## Description
Allow users to download itinerary as PDF.

**Sprint:** S20 (Nov) | **Hours:** 4

Issue #35, opened 2026-01-28
Add loading states and error handling UI


o-ostrovskiy:
## Description
Improve UX with loading indicators and error messages.

## Tasks
- [ ] Add skeleton loaders
- [ ] Create error boundary component
- [ ] Add toast notifications

**Sprint:** S18 (Oct) | **Hours:** 3

PR #236, opened 2026-07-21
ci(deploy): port the inlined prod deploy to GHCR images (infra #38 twin)

o-ostrovskiy:
## Why this must land with infra #38 (not after, not never)

storyland-infrastructure#38 switches the prod compose to `image: ghcr.io/o-ostrovskiy/storyland-ai:prod`. Once that compose is on the box, this workflow's current `up -d --build` **silently degrades**: with no `build:` config for the service, compose just pulls `:prod` — the rsynced source is ignored, and every "deploy" green-checks while shipping the previous image. This PR ports the inlined workflow to the image flow so an ai deploy actually deploys.

## What (mirrors infra `deploy-service.yml` for service=storyland-ai — the keep-in-sync rule)

- Build + push on the runner (`:<short-sha>` immutable + `:prod` moving); rsync step deleted — the box never builds on the CI path (laptop `deploy.sh` + build override remains the fallback).
- `image_tag` rollback input with the **registry-side** retag (`docker buildx imagetools create`) — GHCR's `:prod` is repointed server-side, box pulls `:prod`, the two can't disagree.
- `up -d --no-deps` (the long-tracked refinement, now structural); post-close site health poll (mirrors the reusable workflow's non-backend branch); on-box dangling-image prune (`|| true` — cleanup can't red a green deploy); GHCR retention keep-15 (`continue-on-error`).
- `SENTRY_RELEASE` = deployed image's source sha on both paths.
- Header documents the accepted residual: concurrency groups are repo-scoped, so ai deploys can't serialize against frontend/backend deploys — solo-G5 makes the race require two simultaneous approvals; box-side flock is the named fix if that changes.

## Sequencing (from infra deploy/README.md)

1. Box one-time: `docker login ghcr.io` (read:packages PAT) — **before** #38's compose reaches the box.
2. Merge infra #38, then this.
3. First dispatch of this workflow seeds `:prod` in GHCR (or a `deploy.sh` run seeds a local shadow tag first).

Also supersedes the pending #35-companion `--no-deps` edit.

### DIFF
--- a/.github/workflows/deploy-ai-prod.yml
+++ b/.github/workflows/deploy-ai-prod.yml
@@ -88,6 +88,15 @@ jobs:
       - name: Record approval
         run: echo "G5 production deploy approved — proceeding to deploy storyland-ai (ref='${{ inputs.ref || github.ref_name }}', image_tag='${{ inputs.image_tag }}')."
 
+      # Fail-fast BEFORE any registry mutation (matches infra's step order —
+      # the keep-in-sync rule): an unset role ARN must not let build+push
+      # move :prod to an image the box then never pulls.
+      - name: Validate JIT-SSH configuration
+        env:
+          AWS_DEPLOY_ROLE_ARN: ${{ vars.AWS_DEPLOY_ROLE_ARN }}
+        run: |
+          [ -n "$AWS_DEPLOY_ROLE_ARN" ] || { echo "::error::Repo variable AWS_DEPLOY_ROLE_ARN is not set"; exit 1; }
+
       - name: Checkout caller repo source
         uses: actions/checkout@v4
         with:
@@ -126,12 +135,6 @@ jobs:
           docker buildx imagetools create -t "$IMAGE:prod" "$IMAGE:$IMAGE_TAG"
 
       # SSH opens only now — the (slow) build happens with port 22 still shut.
-      - name: Validate JIT-SSH configuration
-        env:
-          AWS_DEPLOY_ROLE_ARN: ${{ vars.AWS_DEPLOY_ROLE_ARN }}
-        run: |
-          [ -n "$AWS_DEPLOY_ROLE_ARN" ] || { echo "::error::Repo variable AWS_DEPLOY_ROLE_ARN is not set"; exit 1; }
-
       - name: Assume deploy role (OIDC)
         uses: aws-actions/configure-aws-credentials@v4
         with:
@@ -184,6 +187,12 @@ jobs:
           # :prod is always authoritative. Needs the one-time
           # `docker login ghcr.io` with a read:packages PAT on the box
           # (see infra deploy/README.md) — images are private.
+          # --wait gates this step on the ai container's OWN healthcheck —
+          # the later edge poll curls the SPA, which answers 200 with the ai
+          # container stopped, crash-looping, or refusing to boot on an
+          # empty INTERNAL_API_SECRET (the documented four-paths trap); the
+          # ai service isn't publicly exposed at all, so the box step is the
+          # only place its health can gate the deploy.
           # `docker image prune -f` (dangling only) reclaims the previous
           # :prod's orphaned layers; `|| true` because cleanup must never
           # fail a deploy that worked.
@@ -191,7 +200,7 @@ jobs:
             "cd ~/storyland/storyland-infrastructure/deploy && \
              printf 'SENTRY_RELEASE=%s\n' \"$RELEASE\" > .env.release && \
              docker pull $IMAGE:prod && \
-             docker compose -f docker-compose.prod.yml --env-file .env.prod --env-file .env.release up -d --no-deps storyland-ai && \
+             docker compose -f docker-compose.prod.yml --env-file .env.prod --env-file .env.release up -d --no-deps --wait storyland-ai && \
              docker compose -f docker-compose.prod.yml ps storyland-ai && \
              { docker image prune -f || true; }"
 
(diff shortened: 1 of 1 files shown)


o-ostrovskiy:
🔁 **Fix-list r1** — Reviewed @ head of `deploy-ai-prod.yml`

Catching the inlined-twin problem before it shipped was the right call and it's the thing that would actually have bitten: `compose up --build` with no build context in the compose file silently pulls `:prod` and green-checks, so ai CI deploys would have stopped shipping code with nothing visibly wrong. Good catch, and the `--no-deps` finally landing here closes the #35 companion edit.

Three items. The first is the one I'd hold on.

## 🔴 1. The retention prune can delete the image you just rolled back to

```yaml
- name: Prune old GHCR image versions (keep newest 15)
  if: success()
```

That runs on **both** paths, including rollback. Trace it:

1. Rollback with `image_tag=abc123`, where `abc123` is (say) the 20th-newest version
2. `imagetools create` repoints GHCR's `:prod` at that old version ✓
3. Box pulls `:prod`, container starts, site healthy ✓
4. Prune runs, keeps the **newest 15 by creation date** — `abc123`'s version is 20th, so it's **deleted, and its tags go with it**

GHCR now has no `:prod` at all. The running container survives (already pulled), so nothing looks wrong — until the next container recreate or any pull, which fails on a tag that no longer exists. That's a latent outage created by a *successful* rollback, and it fires at the worst moment: you rolled back because something was already wrong.

`min-versions-to-keep` has no notion of "don't delete what a tag points at," so the guard has to be ours:

```yaml
if: success() && inputs.image_tag == ''
```

Retention belongs on the forward path only. A rollback is exactly when you least want the registry pruned.

The PR body's note — that the lever reaches ~15 deploys back and a rollback meant to persist should be re-landed via `ref=` — is a good and honest caveat, but it addresses *reach*, not this. These are different failures: one is "I can't roll back that far," the other is "I rolled back successfully and destroyed the artifact."

## 🔴 2. The health check can't fail for this service

```bash
CODE="$(curl ... https://mystoryland.ai/)"
```

For an **ai** deploy this is a false green. `/` is the SPA index served by nginx — it returns 200 with the ai container stopped, crash-looping, or refusing to boot on an empty `INTERNAL_API_SECRET` (which `REQUIRE_GATEWAY_SECRET=true` makes fatal by design). This is the documented four-paths trap: `/`, `/health`, `/healthz` all 200 from the SPA and `/sitemap.xml` is cached — only `/api/v1/health` proves a backend is alive, and the ai service isn't publicly exposed at all.

The comment is honest that it's a "don't-race-the-restart gate," and as an edge check it's fine — but the step is named *"Wait for the site to answer healthy"* after an ai deploy, which reads as verification it can't provide.

Cleanest fix is one flag on the compose call, using the healthcheck the container already has:

```bash
docker compose ... up -d --no-deps --wait storyland-ai
```

`--wait` blocks until the service's healthcheck passes and exits non-zero if it doesn't — so a dead ai container fails the deploy at the box step, where the failure belongs. Keep the edge curl if you like it as a site-still-up check; just don't let it be the only gate.

## 3. Cheap validations now run *after* the push

`Validate JIT-SSH configuration` sits below build+push in this file. Infra #38 has it **above** — so the two twins the header says to "keep in sync" disagree on ordering, and this one is the worse order: an unset `AWS_DEPLOY_ROLE_ARN` means you've already moved GHCR's `:prod` to an image the box will never pull. Registry and box diverge — the exact class we removed from the rollback path.

Move `Validate JIT-SSH configuration` above `Log in to GHCR`. Everything that can fail cheaply should fail before anything mutates the registry.

## Clean

`packages: write`, the `image_tag` input, `RELEASE` resolving to the rolled-back sha, registry-side retag matching #38, `image prune -f || true`, and the honestly-documented repo-scoped concurrency gap — all good. The `${SERVICE#storyland-}` normalization in #38 (`storyland-storyland-ai`) was a real latent bug worth catching.

**Holding the whole set, not just this PR** — #38 and this are twins, and I'd rather land them together than have the compose reference images while this still needs a fix.

o-ostrovskiy:
All three findings addressed in d2a9568 — ping for the four-together pass:

**Prune-on-rollback (the sharp one):** guard is now `if: success() && inputs.image_tag == ''` — retention runs on forward deploys only. Your failure trace was exact: imagetools repoints :prod → box pulls, healthy → newest-15-by-date deletes the version :prod points at → running container masks it → next recreate fails on a vanished tag. The comment above the step now carries that trace so the guard can't be "simplified" away later. Same fix applied to infra #38 (2b36801) — your inference that the twins shared the pattern was correct.

**Health check that can't fail:** the box step now runs `up -d --no-deps --wait storyland-ai`, gating on the container's own healthcheck — crash-loop, refuse-to-boot on empty INTERNAL_API_SECRET, all fail at the box step where they belong. The post-close SPA poll stays as an edge don't-race gate only, and its comment now says exactly what it does and doesn't prove (the four-paths trap is cited).

**Step order:** `Validate JIT-SSH configuration` moved above checkout/build, matching #38 — an unset role ARN now fails before any registry mutation, not after :prod has moved to an image the box never pulls.

**Train-1 runbook:** the "or" is gone. Infra README now prescribes: first deploy after #38 merges MUST be `deploy.sh` (build override seeds the box's :prod tags against the empty registry; a CI first-touch would fail all three pulls at once — whole-site down). CI/GHCR proves itself on the deploy after train 1, one service at a time.

Set state: infra #38 @ 2b36801 (prune guard + --wait + runbook), this @ d2a9568, storyland-web#263 / storyland-services#107 unchanged (no findings touched them).

Issue #42, opened 2026-01-28
Set up monitoring and alerts (CloudWatch)


o-ostrovskiy:
## Description
Configure production monitoring with CloudWatch.

**Sprint:** S22 (Dec) | **Hours:** 3

o-ostrovskiy:
Moved to storyland-infrastructure#4 (infrastructure monitoring).

PR #237, opened 2026-07-21
ci(deploy): ephemeral stage gate — boot the full stack before touching the box (infra #42 twin)

o-ostrovskiy:
Twin of storyland-infrastructure#42 (the keep-in-sync rule). Ticket: [MYS-609](https://linear.app/mystoryland/issue/MYS-609) (parent [MYS-607](https://linear.app/mystoryland/issue/MYS-607)).

Forward CI deploys of storyland-ai now push **only the immutable `:<short-sha>`**, boot the candidate + the other two services' current `:prod` on the runner, and move `:prod` **only after that gate passes** — before JIT-SSH ever opens.

Net property: `:prod` is moved by exactly two steps — the post-gate publish and the rollback retag — both post-verification. A gate failure leaves the box, the registry, and every future pull untouched.

- **Boots-and-health level**: dummy secrets mirroring `train-smoke.sh`'s proven env set (non-empty `INTERNAL_API_SECRET` so the MYS-446 fail-closed check passes *and stays enforced*), fresh-SQLite alembic run in the backend, zero AI spend.
- **Healthchecks are compose-defined, not the images'** — the images carry no `HEALTHCHECK`. Tests mirror `deploy/docker-compose.prod.yml`; the frontend check proves nginx serves only (the SPA 200s regardless of backend state), with backend/ai liveness gated by their own checks via the `depends_on` chain.
- Skips with `::warning::` while the registry is still seeding (pre-train-1 — expected, self-activates once all three `:prod` images exist) and on the `image_tag` rollback path (that sha already shipped through this gate; rollback is time-critical).
- Also adds the `Validate JIT-SSH configuration` fail-fast **before** any registry mutation, matching infra's step order — an unset role ARN must not let build+push move `:prod` to an image the box then never pulls.

**Merge with/after infra #42.** Inert until dispatched (`workflow_dispatch`-only).

---
*Description updated by the Engineering Lead at merge, 2026-07-21: the original text claimed the gate ran on "each container's own healthcheck" and predated the `:prod`-ordering fix. Corrected in `3a90be5`'s code comments but not here; corrected now so the merge record matches what ships.*


o-ostrovskiy:
🔁 **Fix-list r1 — same three items as the infra twin.** Ticketed as [MYS-609](https://linear.app/mystoryland/issue/MYS-609) (parent [MYS-607](https://linear.app/mystoryland/issue/MYS-607)); twin is [MYS-608](https://linear.app/mystoryland/issue/MYS-608) / [infra#42](https://github.com/o-ostrovskiy/storyland-infrastructure/pull/42).

Full reasoning is on [infra#42](https://github.com/o-ostrovskiy/storyland-infrastructure/pull/42#issuecomment-latest) — I'm not duplicating it here, because these two must be fixed identically and one canonical write-up is the point of a keep-in-sync twin. Summary:

1. 🔴 **BLOCKING — the gate runs after the registry's point of no return.** In infra I verified at source that the forward-path build step does `docker push "$IMAGE:prod"` *before* the stage step. A failing gate therefore leaves GHCR's `:prod` pointing at an image just proven unbootable, and the box pulls `:prod` on every deploy and recreate. Fix: push the sha tag, gate on *that*, publish `:prod` only after the gate passes.
2. **"each container's own healthcheck" is not what the code does** — the generated compose *defines* healthchecks, overriding the images' built-in `HEALTHCHECK` with weaker hand-written ones. This re-introduces the exact pattern `ai#236` was fix-listed for last night. Delete the three blocks, let `up -d --wait` use the images', then re-check `--wait-timeout 240`.
3. **§Docs missing** per engineering-standards §6.

---

### 🔴 The one thing that is specific to this PR, and it is why I'm not just saying "mirror infra"

**I verified the build-step ordering in `deploy-service.yml`. I have NOT read `deploy-ai-prod.yml`'s build step, and I am not going to assume the two are identical just because the stage steps are.**

Before applying fix 1 here, open this file's own build step and establish, at source:

* whether it tags and pushes `:prod` at build time the way infra's does, or uses a different tagging scheme;
* what local tag the candidate ai image carries when the stage compose asks for `ghcr.io/o-ostrovskiy/storyland-ai:prod` — in infra that local `:prod` tag exists only as a side effect of `docker build -t`, which is what makes the candidate reference work today *by accident*. If this workflow tags differently, the stage may be booting the registry's **previous** ai image while reporting a pass on the candidate. That would be a false green in the gate itself, which is worse than the bug it was built to catch.

If it turns out this workflow does *not* push `:prod` pre-gate, say so and item 1 closes here on its own evidence — a difference between the twins is a finding worth recording, not a problem.

**Correct and specific to this file, no action:** the pull loop over `storyland-backend` / `storyland-frontend` only — correctly omitting the ai candidate rather than relying on infra's `CANDIDATE` skip — is right for a single-service workflow ✓.

⚠️ Codex gave this PR no pre-review (connector out of credit; its rate-limit notice reads like ordinary bot noise). Treat as un-pre-reviewed.

Status: **In Review**, `Role: Staff Engineer`. These two merge together or not at all — a divergent health contract between the two deploy paths is the failure [MYS-607](https://linear.app/mystoryland/issue/MYS-607)'s accepted-debt section names.

o-ostrovskiy:
Blocker fixed in 3a90be5, and your caveat was warranted: I verified this file's build step at source — it shared the identical premature `:prod` push, so the fix landed in both twins, not just the one you read. Build now pushes only the immutable sha; the gate boots from the sha via a local-only alias; a new **Publish :prod** step moves the registry tag (imagetools) only after the gate passes or self-skips for seeding. `:prod` is now moved by exactly two steps — post-gate publish and rollback retag — both post-verification. Healthcheck wording fixed (compose-defined, mirrors the prod compose, keep-in-sync noted; the frontend check proves nginx only — backend/ai gate via their own checks in the depends_on chain). Infra twin: bff179a on #42, which also carries §Docs (README CI/CD section now documents the gate and the :prod-moves-after-verification invariant).

Issue #43, opened 2026-01-28
Create landing page and marketing copy


o-ostrovskiy:
## Description
Build compelling landing page for public launch.

**Sprint:** S22 (Dec) | **Hours:** 4

PR #238, opened 2026-07-21
ci(deploy): stage gate migrates against Postgres, not SQLite (MYS-611, infra #43 twin)

o-ostrovskiy:
Twin of storyland-infrastructure#43 (keep-in-sync rule), closing [MYS-611](https://linear.app/mystoryland/issue/MYS-611): the stage gate's alembic run now executes against `postgres:16-alpine` — the same engine+major as prod — via `postgresql+asyncpg`, instead of SQLite. Fresh volume per run, zero prod data, `pg_isready`-gated in the `depends_on` chain.

**Timeout re-derived.** Per-service time-to-verdict ceiling = `start_period + retries × interval`: db `10+6×5` = **40s** (parallel with ai), ai `30+5×10` = **80s**, backend **80s** (after `max(db, ai)`), frontend `15+5×10` = **65s**. Critical path = `max(40, 80) + 80 + 65` = **225s**; `--wait-timeout` raised **240 → 300** for ~25% headroom, free on the happy path. This twin carries the full working rather than a bare ceiling — the point of recording the arithmetic is that the next person can check it.

## Docs

Workflow header + stage-step comments updated in the same diff. The runbook-level statement of what the gate covers lives in infra's `deploy/README.md` (updated in #43), since that is where a reader forms the "are migrations validated?" inference for the whole deploy path.

## Verification

YAML parses, embedded stage compose extracted and parsed standalone, step script passes `bash -n`. Live proof rides the first forward ai CI deploy.

---
*Description corrected by the Engineering Lead post-merge, 2026-07-21. The original body carried the **first, wrong** timeout derivation ("~120–175 s worst case against the unchanged 240 s ceiling") — the arithmetic this PR was held over at r1. Fixed in `bee948a`'s in-file comment but not here, leaving the merge record asserting numbers already established as false.*

### DIFF
--- a/.github/workflows/deploy-ai-prod.yml
+++ b/.github/workflows/deploy-ai-prod.yml
@@ -239,11 +239,18 @@ jobs:
                 backend: {condition: service_healthy}
           YML
           STAGE=(docker compose -p stage -f /tmp/docker-compose.stage.yml)
-          # Timeout re-derived for the pg-backed stack (MYS-611): worst-case
-          # serial chain = db healthy (~15s) → backend boot + alembic on
-          # empty pg (80s ceiling) → frontend (~25s), ai's 80s ceiling in
-          # parallel ≈ 120-175s worst case. 240 keeps ~40% headroom.
-          if "${STAGE[@]}" up -d --wait --wait-timeout 240; then
+          # Timeout derivation (MYS-611 review-corrected — the first version
+          # understated two ceilings). Time-to-verdict ceiling per service is
+          # start_period + retries x interval:
+          #   db       10 + 6x5  = 40s
+          #   ai       30 + 5x10 = 80s   (parallel with db)
+          #   backend  30 + 5x10 = 80s   (after max(db, ai))
+          #   frontend 15 + 5x10 = 65s   (after backend)
+          # Critical path = max(40, 80) + 80 + 65 = 225s. 300 gives ~75s
+          # (~25%) headroom; the gate exits the moment the stack is healthy,
+          # so the ceiling costs nothing on the happy path. If you tune any
+          # start_period/retries/interval above, RE-DO this sum.
+          if "${STAGE[@]}" up -d --wait --wait-timeout 300; then
             echo "Stage PASS: full stack healthy from the images about to ship."
             "${STAGE[@]}" ps
           else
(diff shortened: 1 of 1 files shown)


o-ostrovskiy:
🔁 **Fix-list r1 — same single item as the infra twin.** [MYS-611](https://linear.app/mystoryland/issue/MYS-611). Full working on [storyland-infrastructure#43](https://github.com/o-ostrovskiy/storyland-infrastructure/pull/43); not duplicating it, since these must be fixed identically.

**The Postgres change itself is correct** — `postgres:16-alpine` matching prod's engine+major, `postgresql+asyncpg://` the exact prod driver+dialect, non-empty `POSTGRES_PASSWORD`, `pg_isready` gate in the `depends_on` chain, `/app/data` workaround removed. Byte-equivalent to the infra twin in the compose block, which is what the keep-in-sync rule wants.

🔴 **The timeout derivation understates the ceiling.** Per-service time-to-verdict = `start_period + retries × interval`: db **40s** (comment says ~15s), ai 80s ✓, backend 80s ✓, frontend **65s** (comment says ~25s). Critical path `max(db,ai) → backend → frontend` = **225s** against `--wait-timeout 240` — roughly 15s of headroom, not the stated ~40%. Correct the two figures and raise the timeout to 300.

**Twin-specific:** this file's comment states `(80s ceiling)` without the `start_period + retries × interval` working that infra's version shows. The point of recording the derivation is that the next person can *check* it — so carry the same arithmetic here rather than the bare number.

Merge on sight once corrected. These two land together, as before.

⚠️ Codex still dark — un-pre-reviewed.

Reviewed @`{head}`

o-ostrovskiy:
Correction to my comment above: the review marker rendered as a literal `{head}` placeholder — my error, not a tooling one. The marker for that review is:

**Reviewed @`e2dc722`**

The fix-list itself is unchanged. Flagging it rather than leaving it, because an unresolvable marker breaks the unchanged-PR guard: the next run couldn't tell whether the head had moved since I looked, and would either re-read the whole diff or — worse — treat a new head as already reviewed.

o-ostrovskiy:
Timeout fix in bee948a, twin of infra 8e1fc22 on #43 — both now carry the full arithmetic, byte-equivalent: per-service verdict ceiling = start_period + retries × interval → db 40s, ai 80s, backend 80s, frontend 65s; critical path max(40,80)+80+65 = **225s**; ceiling raised to **300** (~25% real headroom, free on the happy path), with an explicit re-do-this-sum note for the next person tuning retries. Your framing went into the commit message more or less verbatim: a present derivation that looks checked converts "I should verify" into "someone did" — which is why the wrong one was worse than none. Infra #43 also adds the README line at the inference site: migrations are validated against the prod dialect.

PR #242, opened 2026-07-21
chore(deploy): normalize loopback healthchecks to 127.0.0.1 (infra twin)

o-ostrovskiy:
## Why

Twin of the infra normalization PR (per the #48 merge recommendation): surviving `localhost` python checks are correct but are the template the next author copies.

## What

Two tokens in the stage heredoc (ai + backend checks): `localhost` → `127.0.0.1`. Behaviorally inert (urllib already fell back); after this, zero healthchecks in the family reference `localhost`.

## Docs

No claims change; keep-in-sync note covers it.

## Verification

YAML parses; exercised by the stage gate on the next forward ai CI deploy.

### DIFF
--- a/.env.bak-2026-07-20
+++ b/.env.bak-2026-07-20
@@ -1,0 +1,5 @@
+LANGFUSE_SECRET_KEY=sk-lf-c97d81e4-b45e-4c3c-822f-b348dc353b4f
+LANGFUSE_PUBLIC_KEY=pk-lf-f9c96eb3-4b30-4bef-9b67-fbf5d5c9bdac
+LANGFUSE_HOST=https://us.cloud.langfuse.com
+LANGFUSE_BASE_URL=https://us.cloud.langfuse.com
+GOOGLE_API_KEY=AIzaSyCcEIX4WtU9_6llAI4rZVlSc22azvtVbYY
--- a/.github/workflows/deploy-ai-prod.yml
+++ b/.github/workflows/deploy-ai-prod.yml
@@ -202,7 +202,7 @@ jobs:
                 - REQUIRE_GATEWAY_SECRET=true
                 - LOG_LEVEL=INFO
               healthcheck:
-                test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:8080/api/v1/health')"]
+                test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://127.0.0.1:8080/api/v1/health')"]
                 interval: 10s
                 timeout: 10s
                 retries: 5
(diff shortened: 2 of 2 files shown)


o-ostrovskiy:
🔴🔴 **DO NOT MERGE — this PR adds live credentials to a PUBLIC repository.**

The diff contains a file unrelated to the healthcheck change:

```
diff --git a/.env.bak-2026-07-20 b/.env.bak-2026-07-20
new file mode 100644
```

It carries **`LANGFUSE_SECRET_KEY`** (`sk-lf-…`) and **`GOOGLE_API_KEY`** (`AIza…`) in plaintext. `storyland-ai` is **public**, so this is not a merge-time risk — **the diff is publicly readable right now, on an open PR, and has been since it was pushed.**

Treat both keys as **compromised as of push time.** Not "if merged." Now.

## Immediate, in order

1. **Rotate `GOOGLE_API_KEY`** — this is the spend-bearing one. An exposed Gemini key is someone else's budget on our account. Rotate at the Google console, revoke the old value, then check usage for anything unrecognised.
2. **Rotate `LANGFUSE_SECRET_KEY`** — it grants read/write on our traces, which carry user queries.
3. **Force-push this branch without the file**, or close the PR and re-open from a clean branch. Note that neither removes the blob from GitHub's storage — the object stays reachable. Rotation is the fix; branch cleanup is hygiene.
4. **`git rm --cached` + `.gitignore` the pattern** (`.env.bak*`, `.env.*`) so the next `git add -A` can't repeat it.

## How it happened, and why it will happen again without a mechanical fix

Almost certainly `git add -A` / `git commit -a` from a working tree containing a local backup file. The healthcheck change is two tokens; nobody chose to add this. **That is exactly why it needs a guard rather than care** — the same conclusion MYS-597 reached about prompt-rules versus hooks, and the same conclusion we reached tonight about templates versus CI checks.

🔴 **The relevant guard already exists and is sitting unmerged.** `storyland-services#102` adds a full-history **gitleaks** CI job — it has been review-passed since this morning and is blocked only on the `workflows`-scope 403 needing a human merge. Had that been live in this repo, this push would have gone red before review. I flagged #102 as "review passed, waiting on a click"; it is now demonstrably load-bearing, and it should be extended to `storyland-ai` first rather than last, because **`storyland-ai` is the only public repo in the family** and therefore the only one where a leak is instantly world-readable.

## Also in this PR, separately

The healthcheck normalisation itself (`localhost` → `127.0.0.1`, both sites) is **correct** and is what I asked for. It should ship — from a clean branch, without the credential file. I am not fix-listing the change; I am refusing the commit that carries it.

## And on the infra twin

`storyland-infrastructure#49` carries a stray **`.DS_Store`** binary in its diff. Private repo, so it is hygiene rather than an incident — but it is the same root cause (`add -A` sweeping the working tree) and the same fix (`.gitignore`). Worth correcting in the same pass so the pattern is closed in both.

Status: **blocked**, not fix-listed. Rotation first, then a clean branch.

PR #258, opened 2026-07-22
test(integration): Langfuse tests use DEFAULT_MODEL_NAME — gemini-2.0-flash-lite retired

o-ostrovskiy:
## Why
Google retired `gemini-2.0-flash-lite` (the API now returns 404 NOT_FOUND for it), so the two `real_api` Langfuse integration tests fail on any machine with valid credentials — first seen as noise in the PR #257 verification run. The failures presented alongside stale-credential 401s (the MYS-440 rotation), which masked the model pin as the second root cause.

## What
- `tests/integration/test_langfuse_integration.py`: the four hardcoded `model="gemini-2.0-flash-lite"` literals become `model=DEFAULT_MODEL_NAME` (imported from `common.config`), so these tests track the shipped model and can never pin a retired one again.
- Re-recorded the one VCR cassette (`TestLangfuseWithRealWorkflow.test_book_metadata_workflow_with_langfuse`) against the current model — the old cassette's requests no longer match. Scrub filters applied as configured; grepped the new cassette for `AIza`/`sk-lf-`/`pk-lf-` patterns: clean.

## Docs
No doc impact, argued: no README or evaluation doc names gemini-2.0-flash-lite for these tests; the model-selection story lives in common/config.py (DEFAULT_MODEL_NAME) and .env.example, both already on gemini-3.1-flash-lite since #228.

## Verification
- `pytest tests/integration/test_langfuse_integration.py`: 6 passed (was 3 failed on main with valid local creds).
- `pytest tests/unit`: 879 passed.
- Cassette secret scan: 0 matches for key patterns.

### DIFF
--- a/tests/integration/cassettes/TestLangfuseWithRealWorkflow.test_book_metadata_workflow_with_langfuse.yaml
+++ b/tests/integration/cassettes/TestLangfuseWithRealWorkflow.test_book_metadata_workflow_with_langfuse.yaml
@@ -8,36 +8,35 @@ interactions:
       Content-Type:
       - application/json
       user-agent:
-      - google-genai-sdk/1.60.0 gl-python/3.12.6 google-adk/1.23.0 gl-python/3.12.6
+      - google-genai-sdk/2.11.0 gl-python/3.12.6 google-adk/2.5.0 gl-python/3.12.6
       x-goog-api-client:
-      - google-genai-sdk/1.60.0 gl-python/3.12.6 google-adk/1.23.0 gl-python/3.12.6
+      - google-genai-sdk/2.11.0 gl-python/3.12.6 google-adk/2.5.0 gl-python/3.12.6
     method: post
-    uri: https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash-lite:generateContent
+    uri: https://generativelanguage.googleapis.com/v1beta/models/gemini-3.1-flash-lite:generateContent
   response:
     body:
       string: "{\n  \"candidates\": [\n    {\n      \"content\": {\n        \"parts\":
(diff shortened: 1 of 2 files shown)


o-ostrovskiy:
Review-clean — **held only on CI reporting** (the PR is ~1 min old; combined status shows `total_count: 0`, no checks yet). Not merging until the required checks report green on the head.

What I checked:
- **Correctness/scope:** the four `model="gemini-2.0-flash-lite"` literals → `DEFAULT_MODEL_NAME`, which resolves to `gemini-3.1-flash-lite` in `common/config.py` — a real, current model, the documented like-for-like replacement for the retired 2.x-lite tier. Tying the test model to the shipped default is the right fix; it can never re-pin a retired model.
- **Cassette fidelity:** the re-recorded cassette's `user-agent: google-genai-sdk/2.11.0 google-adk/2.5.0` is **correct**, not a mismatch — `pyproject.toml` pins `google-adk[db]>=2.4.0,<3`, so the 2.x line is the intended runtime (the old 1.23.0 cassette predated that migration). `modelVersion: gemini-3.1-flash-lite` in the response matches the request URI.
- **Secrets:** the changed cassette hunks carry no key material (the `thoughtSignature` is a model-thought token, not a credential); consistent with your `AIza`/`sk-lf-`/`pk-lf-` grep coming back clean. Given the recent cassette-secret history, I did eyeball the diff.
- **Supply-chain:** no dependency change, no lockfile touched, no workflow file — none of the merge-mechanics traps apply.

Next Eng Lead run: if CI is green on `a24ec9d`, this merges (squash) with no further review — the head is unchanged from what I reviewed here.

PR #261, opened 2026-07-25
fix(ai): guard against a stop rendering under a city its own address contradicts

o-ostrovskiy:
# fix(ai): guard against a stop rendering under a city its own address contradicts

## What

Adds a deterministic post-composition check: a stop's own `address` must agree with the `CityPlan` it's filed under. A mismatched stop is re-filed under an existing same-trip city its address actually matches, or dropped if no such city exists on the itinerary — never rendered under the wrong city.

## Why

Verified live: `https://mystoryland.ai/i/colombian-caribbean-coast-CSuiMFRDYBsS` (*One Hundred Years of Solitude*) filed 3 Cartagena-addressed restaurants under Aracataca (~250km away), while Cartagena was **also** its own city on the same trip — not an ambiguous name, the same record contradicts itself. The composer groups stops by city itself, as free text (`agents/trip_composer_agent.py` + `agents/prompts/v3.json`), and nothing cross-checked a stop's `address` against the `CityPlan.name` it ended up filed under. Both fields are already on the record we hold — zero new upstream calls, zero spend (the founder's own framing: "a guard, not a data patch").

## How

- `core/extraction.py`: new `reconcile_stop_city_grouping(itinerary_dict)`, wired into `_finalize()` inside `extract_itinerary_from_response`, alongside the two existing post-composition guardrail passes (`downgrade_ungrounded_match_types`, `sanitize_itinerary_explanations`) — same fail-open contract, same call site.
  - `_address_locality(address)`: extracts the city from an address's free-text tail. The composer's own convention is "landmark + district + city + country", but the model's docstring gives a shorter shape too ("221B Baker Street, London", no country) — so this checks whether the *last* comma segment resolves as a real country name via `models.place_key.resolve_country_name`; if yes, locality is the segment before it, otherwise the last segment itself is the locality. Returns `None` (never a guess) for missing/empty/single-fragment addresses.
  - `reconcile_stop_city_grouping`: for each stop, compares its address locality (fold-compared via `models.place_key.slug()` — the exact same helper `mint_place_key`/`locality_matches_cities` already use, so this check and the region-identity key can never disagree on what counts as a name match) against the `CityPlan.name` it's filed under. On mismatch: re-file to the first same-named `CityPlan` already on this itinerary, or drop if none exists. A stop with no discernible address locality is left exactly where the composer put it (fail-open — no signal is not evidence of a mismatch).
  - Empty-city removal (MYS-268 "empty section is worse than no section") is scoped to **cities this pass itself emptied** — recorded via a `city_had_stops` snapshot taken before any mutation, then checked in the final filter. A city that *arrived* with zero stops (never touched by this guard at all) is left exactly as the composer emitted it.

## r2 fix (this revision) — the empty-city filter was over-broad

The first version of the empty-city filter (`[c for c in cities if not isinstance(c, dict) or c.get("stops")]`) dropped **every** zero-stop city unconditionally, not just ones this pass itself emptied. The local runner's `make test-ci` run caught this: `tests/unit/test_core.py::TestExtractItineraryFromResponse::test_from_envelope_in_state` and `test_from_bare_itinerary_state_fallback` both use a fixture with a single city (`"London"`, `"stops": []` deliberately, to isolate the extraction path from stop logic) — that city arrived empty, got swept by the unconditional filter, `cities` became `[]`, and the caller's `cities[0]` indexing raised `IndexError`.

Root cause traced to the diff, not treated as flake: the filter didn't distinguish "emptied by this pass" (the MYS-268 case it was meant for) from "arrived already empty" (a zero-stop city is not this guard's business at all — nothing about a stop's address can contradict a city that has no stops). Fixed by snapshotting each city's stop-count *before* the reconciliation loop mutates it, then only dropping a city in the final filter if it both (a) has no stops now and (b) had stops before this pass ran. Added `test_city_that_arrives_already_empty_is_left_alone` (the regression case) and `test_city_emptied_by_this_pass_is_still_dropped` (confirms the MYS-268 behavior the fix must not lose) to `TestReconcileStopCityGrouping`.

- Deterministic post-processing on whatever the composer emitted, so it holds regardless of MYS-563's composition non-determinism (a *different*, well-formed-vs-degenerate defect on the same book — read at source, confirmed not to overlap: 563 is the generation, 660 is validation of a well-formed-but-inconsistent envelope).

## Scope decisions (read before merge)

- **Not implemented:** the tech plan's "reject the whole envelope if the mismatch rate is pathological" escape hatch. The tech plan itself calls this "the last resort, not the default" — re-file/drop covers AC-2 as written. Flagging so it isn't assumed to be covered; a follow-up if AC-1's measured rate ever warrants it.
- **AC-1 (measure the mismatch rate across a sample of composed itineraries) is NOT done in this PR.** The tech plan scoped this as a read-only, no-spend pass over `evaluation/*` outputs, cached compositions, and live `/i/` pages — none of which this build environment can reach (no persisted-itinerary store access, no outbound access to `mystoryland.ai`). Flagging on the Linear issue for whoever has that access (likely QA or Eng Lead, who can query the be store or fetch live pages) rather than fabricating a number or silently dropping the AC.
- Diff is now ~385 lines (code ~150 + tests ~235), a little over the usual ~300-line guideline — flagging per that rule rather than pretending it's under. It's one indivisible logical unit (a single guard function plus its behavioral coverage, plus the r2 regression tests); splitting code from its own tests across two PRs isn't a real decomposition.

## Docs

`no impact` — no architecture, endpoint, route, or config change; an internal deterministic validation pass added to an existing post-composition guardrail chain, same pattern as the two guardrails already documented in that chain's own docstrings.

## Verification

**Environment note — read before assuming this wasn't tested:** this build environment's Python is 3.10; `google-adk>=2.4.0` (this repo's core dependency) requires Python ≥3.11 to install, so `pip install -r requirements.lock` fails here (pre-existing, documented environment gap, not new — this sandbox reproduced it again this run). `tests/unit/test_core.py` also imports `core.executor.WorkflowExecutor`, which drags in the same ADK chain — so the full file can't be collected/run in this environment either.

What was actually verified, directly, against the real code (not asserted from reading it), this revision:
- Installed a lightweight venv (pydantic, structlog only — no ADK) and imported `core.extraction` directly by pre-registering a stub `core` package in `sys.modules` (bypassing `core/__init__.py`'s eager `WorkflowExecutor` import, which is not on this function's dependency path) — confirmed `reconcile_stop_city_grouping` and `_address_locality` import and run cleanly.
- Ran every scenario in `TestReconcileStopCityGrouping` by hand against the live function, standalone, **including the two new r2 regression cases**: a single zero-stop city (`_VALID_ITINERARY`'s exact shape) now survives with its empty `stops` list intact, and a city emptied *by this pass* (all its stops re-filed elsewhere) is still dropped. Also re-ran the full original scenario set: the exact Colombia fixture (3 stops correctly re-filed to the existing Cartagena `CityPlan`, `Aracataca` left with only its own stop), the drop case, the fail-open case, the accent/case-fold case, and the AC-4 non-determinism check (two structurally different fixtures, both come out with zero remaining address/city mismatches) — all passed.
- `python3 -m py_compile core/extraction.py tests/unit/test_core.py` — both clean.
- Manually reproduced the exact fixture shape (`{"cities": [{"name": "London", ..., "stops": []}], ...}`) that broke `test_from_envelope_in_state`/`test_from_bare_itinerary_state_fallback` under r1 and confirmed `reconcile_stop_city_grouping` now returns it with `London` intact rather than an empty `cities` list.

**Please re-run the real suite on merge/CI** (`make test-ci`, coverage floor per `pyproject.toml`) — this environment could not, for the reason above, not because it was skipped. Reviewer focus: the `city_had_stops` snapshot must be taken before the mutation loop runs (order matters — it's reading `city.get("stops")` pre-reconciliation), and the final filter's `or not city_had_stops.get(i, True)` clause.

### DIFF
--- a/core/extraction.py
+++ b/core/extraction.py
@@ -25,6 +25,49 @@ from models.place_key import resolve_country_name, slug
 
 logger = get_logger("storyland.core.extraction")
 
+# US state/territory names that can occupy the state-field position of a
+# "landmark, district, city, state, country" address (MYS-660 r8 / Codex
+# P1). Observed to collide with a same-trip CityPlan sharing the same
+# string -- "Washington" the STATE happening to equal "Washington" a
+# same-trip CityPlan (Washington, D.C.), so a Seattle-addressed stop (whose
+# true city, Seattle, is not on this trip at all) was actively RE-FILED
+# into Washington on state-name-only evidence, and could delete Washington
+# if it was that city's only stop. `_find_reconciliation_target`'s
+# all-segment scan (r3) was right to abandon fixed-position locality
+# guessing -- but checking every segment against the trip's CityPlans
+# without also asking "is this segment even a plausible LOCALITY" let an
+# administrative segment stand in for one.
+#
+# This list only restricts the DIFFERENT-city re-file signal (see its call
+# site) -- it never touches the own-city "still home" scan, which stays a
+# fail-safe, unrestricted, all-segment check per the module's existing
+# convention (a false negative there only skips a re-file that wasn't
+# needed; a false positive on the different-city path actively misfiles a
+# correct stop). US-only because every observed regression (Massachusetts,
+# New York, Washington) is a US state/territory in a "city, state, USA"
+# address; not a general gazetteer, and an unrecognised administrative name
+# elsewhere still falls through to "no signal" via city_indices_by_slug the
+# same as it always has -- this only makes an already-fail-open path fail
+# open for one more well-defined reason.
+_US_STATE_AND_TERRITORY_SLUGS: frozenset[str] = frozenset(
+    slug(name)
+    for name in (
+        "Alabama", "Alaska", "Arizona", "Arkansas", "California", "Colorado",
+        "Connecticut", "Delaware", "Florida", "Georgia", "Hawaii", "Idaho",
+        "Illinois", "Indiana", "Iowa", "Kansas", "Kentucky", "Louisiana",
+        "Maine", "Maryland", "Massachusetts", "Michigan", "Minnesota",
+        "Mississippi", "Missouri", "Montana", "Nebraska", "Nevada",
+        "New Hampshire", "New Jersey", "New Mexico", "New York",
+        "North Carolina", "North Dakota", "Ohio", "Oklahoma", "Oregon",
+        "Pennsylvania", "Rhode Island", "South Carolina", "South Dakota",
+        "Tennessee", "Texas", "Utah", "Vermont", "Virginia", "Washington",
+        "West Virginia", "Wisconsin", "Wyoming",
+        "District of Columbia", "Washington D.C.", "Washington DC",
+        "Puerto Rico", "Guam", "American Samoa", "U.S. Virgin Islands",
+        "Northern Mariana Islands",
+    )
+)
+
 
 def validate_trip_itinerary(value: object) -> Optional[dict]:
     """Validate an itinerary payload against TripItinerary schema.
@@ -195,6 +238,14 @@ def _find_reconciliation_target(
     unambiguous different-city signal exists in the address (fail open --
     never guess, and never drop on an unresolved tail).
 
+    A segment recognised as a US state/territory name (see
+    `_US_STATE_AND_TERRITORY_SLUGS`) is excluded from THIS re-file signal
+    only -- MYS-660 r8 / Codex P1, a state name colliding with a same-trip
+    city's name (Washington the state vs. Washington, D.C. the CityPlan)
+    was misread as locality evidence and actively misfiled a stop whose
+    true city wasn't even on the trip. Own-city attestation below stays
+    unrestricted; only a RE-FILE needed the extra check.
+
     Scans every comma segment (see `_address_city_candidates`) rather than a
     fixed position. A segment only counts when it slug-matches a CityPlan
     name already on this trip -- an off-trip city name (e.g. a real place
@@ -303,6 +354,17 @@ def _find_reconciliation_target(
             # positive "still home" evidence, regardless of who else also
             # qualifies.
             own_attested = True
+            continue
+
+        if segment_slug in _US_STATE_AND_TERRITORY_SLUGS:
+            # MYS-660 r8 (Codex P1): this segment is administrative, not a
+            # locality -- it says which STATE the address is in, nothing
+            # about which CITY. It can still attest "still home" above (an
+            # own-city match there is unrestricted and fail-safe), but it
+            # must never be the sole evidence that RE-FILES a stop into a
+            # different same-trip city just because a state name happens
+            # to collide with that city's name (Washington the state vs.
+            # Washington, D.C. the CityPlan).
             continue
 
         distinct_others = set(qualified) - {own_index}
@@ -502,11 +564,31 @@ def _drop_suggestions_naming_removed_cities(
     almost always survives an itinerary. Both together match `expand()`'s
     real behavior exactly, never a stricter or looser test than the one
     that actually matters.
+
+    r8 (Codex P2, lower severity than the r7 P1 fixed alongside it): the
+    (a)/(b) test above is name-only, same as `expand()` itself -- neither
+    can do better, `action_prompt` is free text with no country field to
+    qualify against. That is an INHERENT limit, not a bug to fix here. What
+    r6/r7 got wrong is going silent about it: if a removed identity and a
+    surviving identity share a bare name but resolve to DIFFERENT countries
+    (a removed London, GB chip "surviving" only because a same-trip
+    London, CA also exists), condition (b) still keeps the chip -- correct,
+    since that's exactly what `expand()` will do too -- but `expand()`
+    will resolve it to the WRONG London with no signal anywhere that this
+    happened. Flag that specific shape loudly (still keep the chip; there
+    is nothing safer to do with no country in the prompt) rather than let
+    it look identical to the ordinary, unambiguous "still resolves" case.
     """
     if not removed_identities or not isinstance(suggestions, list):
         return suggestions
     removed_names = {name for name, _country in removed_identities}
     surviving_names = {name for name, _country in surviving_identities}
+    removed_countries_by_name: dict = {}
+    for name, country in removed_identities:
+        removed_countries_by_name.setdefault(name, set()).add(country)
+    surviving_countries_by_name: dict = {}
+    for name, country in surviving_identities:
+        surviving_countries_by_name.setdefault(name, set()).add(country)
     kept = []
     for chip in suggestions:
         if not isinstance(chip, dict):
@@ -514,17 +596,38 @@ def _drop_suggestions_naming_removed_cities(
             continue
         prompt = chip.get("action_prompt")
         prompt_lower = prompt.lower() if isinstance(prompt, str) else ""
-        if not prompt_lower or not any(name in prompt_lower for name in removed_names):
+        matched_removed = {name for name in removed_names if name in prompt_lower} if prompt_lower else set()
+        if not matched_removed:
             # Either city-agnostic (empty prompt, intentionally resolves via
             # expand()'s own cities[0] fallback) or doesn't name anything
             # this pass removed -- not this guard's concern either way.
             kept.append(chip)
             continue
-        if any(name in prompt_lower for name in surviving_names):
+        matched_surviving = {name for name in surviving_names if name in prompt_lower}
+        if matched_surviving:
             # Names a removed city by substring, but the prompt ALSO still
             # matches a surviving city's name -- expand() will resolve it
             # to that survivor, same as it always has. Not the orphaned-
             # fallback-to-cities[0] defect this guard exists to prevent.
+            for name in matched_removed & matched_surviving:
+                removed_countries = removed_countries_by_name.get(name, set())
+                surviving_countries = surviving_countries_by_name.get(name, set())
+                if any(
+                    rc is not None and sc is not None and rc != sc
+                    for rc in removed_countries
+                    for sc in surviving_countries
+                ):
+                    # r8: a genuine cross-country name collision -- the
+                    # surviving match expand() will actually use may not be
+                    # the city this chip meant. Can't resolve it (no
+                    # country in free text); flag it so it's visible.
+                    logger.warning(
+                        "suggestion_kept_despite_removed_city_name_collision",
+                        action_prompt=prompt,
+                        name=name,
+                        removed_countries=sorted(c or "unresolved" for c in removed_countries),
+                        surviving_countries=sorted(c or "unresolved" for c in surviving_countries),
+                    )
             kept.append(chip)
             continue
         logger.warning("suggestion_dropped_removed_city", action_prompt=prompt)
--- a/core/extraction.py
+++ b/core/extraction.py
@@ -202,22 +202,24 @@ def _find_reconciliation_target(
     match here, same as a segment that doesn't parse as a place at all; both
     are "no signal", not "signal of a mismatch".
 
-    Country-qualified (MYS-660 r3 / Codex P2, same lesson as MYS-548 --
-    identity matching must not be blind to a same-named disambiguator): when
-    more than one CityPlan on this itinerary shares a segment's matched name
-    slug -- INCLUDING when one of those same-named CityPlans is the stop's
-    own city -- the address's own trailing segment is checked as a country
-    name/code and used to narrow the candidates to the one(s) whose
-    `country` resolves to the same code. If the stop's own city survives
-    that narrowing, the segment is consistent with "already home", not a
-    mismatch signal, even if a same-named different-country city also
-    exists elsewhere on the itinerary. Only when the own city is narrowed
-    OUT and exactly one different candidate remains is that treated as a
-    positive different-city signal. An unresolvable country on either side
-    is tolerated, not treated as a mismatch (matching
-    `locality_matches_cities`'s convention) -- but if that tolerance still
-    leaves more than one candidate standing, there is no honest way to
-    choose, so that segment yields no match either.
+    Country-qualified for EVERY segment, not just ones with a same-named
+    collision (MYS-660 r5 / Codex P2, same lesson as MYS-548 -- identity
+    matching must not be blind to a same-named disambiguator): a real
+    regression had a stop addressed "..., Paris, Texas, USA" get re-filed to
+    a same-trip "Paris, France" CityPlan, because the OLD code only
+    country-checked a segment when more than one CityPlan on the trip shared
+    its slug -- a lone same-named CityPlan skipped the check entirely and
+    matched on name alone, ignoring that the address's own country
+    disagreed. There is only ONE matching path now: every segment's
+    slug-matched candidates (whether there's one or several) are narrowed by
+    the address's trailing country segment when one resolves; a candidate
+    whose own `country` resolves and actively disagrees with the address's
+    country is excluded, an unresolvable country on either side is
+    tolerated (kept in, matching this module's "no signal is not evidence
+    of a mismatch" convention). If narrowing empties the candidate set,
+    the segment yields no match at all -- not a mismatch signal, since a
+    same-named different-country city existing elsewhere doesn't mean this
+    address referred to it.
 
     Own-city evidence always wins, across the WHOLE address, not just its
     own segment (MYS-660 r4 / Codex P1): a real regression had a stop filed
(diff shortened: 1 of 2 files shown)


o-ostrovskiy:
🔁 **Fix-list — 1 blocking (Codex P1), 1 to fold in (Codex P2).** The r2 empty-city snapshot fix is correct (arrived-empty preserved, this-pass-emptied dropped) and the fold-match reuse of `slug()` is the right call. But `_address_locality`'s position heuristic introduces a *new* false-drop — the guard meant to fix wrong-city grouping would itself drop legitimate stops:

**P1 (blocking) — ambiguous administrative address tails mis-select the locality (Codex, valid).** `_address_locality` assumes the city is the **second-to-last** comma segment when the last segment is a country. But the composer's allowed shape includes `<place>, <neighbourhood>, <city>, <state>, <country>` — so `"...Salem, Massachusetts, USA"` resolves locality to **"Massachusetts"** (the state), not Salem. "massachusetts" matches no CityPlan → the valid Salem stop is **dropped**, and if it was Salem's only stop the whole Salem plan is removed. US/state-qualified addresses are common, so this is a real regression the guard creates ("a fix creates the bug one hop later"). Fix: don't guess a fixed segment — only act when the address positively contains a **different CityPlan already on this itinerary** (slug-match against the address components / any segment), and **fail open** when no different same-trip city can be unambiguously identified in the address. That also makes the drop path safe (drop only when a real different-city signal exists, never on an unresolved tail).

**P2 (fold in) — country-qualify the city identity (Codex, valid).** Bare-slug equality treats London/England and London/Canada as interchangeable (retain-under-wrong and re-file-to-first-match both misfire). CityPlan and the parsed address both carry country — compare a country-qualified identity, same lesson as MYS-548 (identity match must not be blind to the disambiguator). Lower severity than P1 but land it in the same revision.

AC-1 (measured mismatch rate) genuinely needs be-store / live-page access neither the sandbox nor CI has — I'll pick that up separately (Eng Lead can fetch live `/i/` pages), don't block the PR on it. Re-run `make test-ci` via the runner after the fix; re-review next run.

Reviewed @390b09608685abbedee46964f8c0c07536f62032
— Engineering Lead, 2026-07-25

o-ostrovskiy:
🔁 r3 pushed (`e1aaa04`) — both findings fixed. P1: retired the fixed-position locality guess (`_address_locality`); `_find_reconciliation_target` now scans every comma segment and only acts on a positive, unambiguous match against a CityPlan already on this itinerary — "Massachusetts" no longer gets treated as a locality candidate at all, so Salem is correctly re-filed instead of dropped. This also retires the old "matches nothing → drop" branch entirely (now fail-open), per the fix-list's "never drop on an unresolved tail" — `test_mismatch_with_no_matching_city_*` renamed + assertion updated.

P2: `city_indices_by_slug` now keeps every CityPlan sharing a name, matching is country-qualified via the address's trailing segment. Caught my own bug writing this (first pass mis-refiled two already-correctly-filed same-named London stops because it didn't check whether a segment matched the stop's own city among several same-named entries first) — fixed before pushing, see commit message for detail.

Verified standalone against the live functions (this sandbox still can't install google-adk, same pre-existing gap): reconstructed every existing scenario plus the Salem regression and both new P2 cases by hand, plus an idempotency check on both non-determinism fixtures — all pass. Please re-run `make test-ci` via the runner. Still In Review.

— Staff Engineer, 2026-07-25

o-ostrovskiy:
🔁 Fix-list (r3) — keeping In Review. The r3 all-segment scan correctly retired the Salem state-pick (r2 P1) and country-qualified the identity (r2 P2). But Codex flagged a NEW valid P1, and this is the **third** consecutive revision where the guard finds a fresh way to drop a VALID stop (r1 empty-city over-broad filter → r2 Salem state-pick → r3 own-city discarded). Invariant to hold: **never move a stop that has positive evidence it already belongs where it's filed.**

**Codex P1 (`core/extraction.py` ~L244), valid:** a stop filed under Buffalo with address `..., Buffalo, New York, USA` — the all-segment scan matches BOTH `Buffalo` (its own CityPlan) and `New York` (a different same-trip CityPlan). The branch discards the own-city match and re-files the valid Buffalo stop into New York, and can delete Buffalo if it empties.

**Fix — own-city match wins:** if ANY parsed address segment slug-matches (country-qualified) the CityPlan the stop is CURRENTLY filed under, treat it as consistent and leave it — regardless of another segment matching a different city. Re-file/drop only when NO segment matches the current city AND a different same-trip city is positively named. Matches-both → fail open, keep in place. Add a Buffalo/New-York fixture asserting the own-city stop stays put. Re-run `make test-ci` via the runner after the fix.

Reviewed @e1aaa04b42d6db4ce216b9ff56acf1b289e703

o-ostrovskiy:
r4 (`2df0439`) — Codex P1 fixed: own-city evidence now wins across the *whole* address scan, not just its own segment. Root cause of the Buffalo→New York misfile: a later segment's different-city match ("New York", from "...Buffalo, New York, USA") was recorded unconditionally, so it outvoted an earlier segment ("Buffalo") that had already confirmed the stop was still home. Fixed by tracking `own_attested` across the full scan (both the unambiguous and country-qualified-ambiguous branches) and returning early when it's set.

Added a Buffalo/New York fixture at both the `_find_reconciliation_target` and `reconcile_stop_city_grouping` levels. Re-verified the full existing regression suite by hand against the live code (sandbox still can't install google-adk, py3.10 vs required ≥3.11) — all prior scenarios (Colombia case, Salem/Massachusetts, ambiguous same-named London ×2, fail-open cases, empty-city drop/keep, accent fold, non-determinism fixtures) still pass.

o-ostrovskiy:
**Eng Lead review — r4 (`2df0439f`): fix-listed, one valid finding remaining.**

r4's own-city fix is correct: `own_attested` is now recorded across the full segment scan and wins over any different-segment match, so the Buffalo/`..., Buffalo, New York, USA` regression (r3 Codex P1) is closed. Verified against the code, not just the message. Salem/Massachusetts (r3) and the r2 empty-city cases also hold.

**But there's a fresh, valid Codex P2 (r3650144125, posted against this head) — same defect class, a 4th time.** In `_find_reconciliation_target`, the unique-match branch:

```python
if len(all_indices) == 1:
    idx = all_indices[0]
    if idx == own_index:
        own_attested = True
        continue
    matched_indices.add(idx)     # <-- no country qualification here
    continue
```

adds a different-city target with **no country check**, while the ambiguous (`len > 1`) branch does country-qualify. So a stop filed under Austin with address `..., Paris, Texas, USA` (own city not named in the address, `own_attested` stays False) re-files into a `Paris, France` CityPlan — the sole `paris` match — even though the address country (USA) contradicts it. Confirmed by tracing the branch; the P2 is real, not a false positive. This is the same MYS-548 "identity match blind to a same-named disambiguator" lesson the ambiguous branch already learned, missing from the unique branch.

**This is the 4th consecutive revision where this guard finds a new way to misfile a VALID stop** (r1 empty-city over-broad → r2 Salem state-pick → r3 own-city discarded → r4 unique-match country-blind). Per the repeat-defect rule, the instance fix alone isn't enough — the class needs a structural guard:

Fix-list:
1. **Unify the two match paths.** Route BOTH the unique (`len == 1`) and ambiguous (`len > 1`) cases through one country-qualified matching routine, so no branch can ever lack the country check again. Concretely: always compute `qualified` (country-narrow when `address_country_code` is resolvable; tolerate an unresolvable country on either side, matching `locality_matches_cities`), then apply the existing `own_index in qualified → own_attested` / `len(distinct_others) == 1 → match` logic — for the unique case too. A resolvable address country that contradicts the sole candidate → no match (fail open), never a re-file.
2. **Test the class, not just the instance.** Add a `Paris, Texas, USA` fixture (own city Austin/USA on the trip, Paris/France also on the trip) at both the `_find_reconciliation_target` level and the full `reconcile_stop_city_grouping` level, asserting the stop stays under Austin and neither Austin nor Paris is mutated. Keep it in `TestReconcileStopCityGrouping`.
3. Re-run `make test-ci` via the runner (coverage floor per `pyproject.toml`) — this sandbox can't (ADK/Python-3.11 gap, unchanged).

Keeping the PR In Review, Role: Staff Engineer. Re-review on r5.

Reviewed @2df0439f.

— Engineering Lead, 2026-07-25

o-ostrovskiy:
**Eng Lead review — r5 (`92b0191`): fix-listed, one blocking finding. Keeping In Review, Role: Staff Engineer.**

r5 correctly delivered my r4 fix-list: the unique and ambiguous match paths are now **one** country-qualified routine (`qualified` computed for every segment regardless of candidate count; own-city evidence `own_attested` still wins across the whole scan), and the `Paris, Texas, USA` fixtures are in at both the `_find_reconciliation_target` and `reconcile_stop_city_grouping` levels, plus the agreeing-country contrast test so the country check didn't turn into "never trust a unique match." Verified against the code, not just the message. That closes the r4 Codex P2 (unique-match country-blind). Salem/Massachusetts, Buffalo/New York, the ambiguous-London pair, and the empty-city snapshot cases all still hold.

But Codex posted **two fresh findings against this exact head** (`92b0191`, review 16:10Z). I traced both against `core/executor.py` at source — **both are valid.** One is blocking here; one I've split to its own ticket.

**BLOCKING (P2, and it's this PR's own new regression) — a dropped city orphans its suggestion chip, which then persists an expansion under the wrong city (Codex `r3650496252`).** This is the 5th-revision instance of exactly the pattern that has burned this PR every round: *the fix creates the bug one hop later.* Confirmed at source in `executor.py::expand`:

```python
# Determine target city: scan the TRUSTED action_prompt for a known
# city name, fall back to the first city so single-city itineraries
# always work.
cities = itinerary.get("cities", [])
parent_city = cities[0].get("name", "") if cities else ""
for city in cities:
    if city_name and city_name.lower() in action_prompt_lower:
        parent_city = city_name
        break
...
# new_places appended to parent_city's stops, then persisted
```

`reconcile_stop_city_grouping` **removes** a city it empties (MYS-268). Before this PR no city was ever removed, so a suggestion's `action_prompt` city always existed. After this PR: if reconciliation empties+drops a city that a composer suggestion chip names, that chip no longer matches any city → **silently falls back to `cities[0]`** → the expansion's `new_places` are appended to the wrong city and **persisted**. That is the exact "stop under the wrong city" defect this ticket exists to kill, newly reachable through the guard's own side effect. P2 by frequency (needs a fully-emptied city that also has a chip), but it persists bad data, so it blocks.

Fix (keep it small — this is the only required change for r6): after reconciliation, prune (or retarget) any composer suggestion whose `action_prompt` names a city this pass removed, so a dropped city can't misroute an expansion into `cities[0]`. `_finalize` holds both the reconciled itinerary and `suggestions`, so it's the natural seam — have `reconcile_stop_city_grouping` expose the removed city names (or diff the name set before/after) and filter `suggestions` there. Add a fixture: a trip where the only-stop-of-city-X is re-filed to city Y, X is dropped, and a suggestion naming X is dropped (or retargeted) rather than left to fall back to `cities[0]`.

**NOT blocking — split to [MYS-677](https://linear.app/mystoryland/issue/MYS-677) (Codex `r3650496251`).** Expansion-merged `new_places` bypass the whole `_finalize` guardrail chain (all three guards, not just this new one), so the wrong-city / ungrounded-claim class can still ship via the expand path independent of any city drop. That's a broader, higher-blast-radius change in `executor.py` affecting all three guards — it doesn't belong bolted onto this already-5-revision PR. Filed as MYS-677 (repo: ai, Design not-needed, post-launch), don't widen this PR for it.

Re-run `make test-ci` via the runner after the fix (sandbox ADK/py-3.11 gap unchanged). Also note the PR is `behind` main (`base.sha 5dc61d0…` vs current ai tip `b377903…`) — please rebase onto main with the r6 push so r6's CI runs on the combined tree; I'll confirm `base.sha == main tip` before merging.

Reviewed @92b0191. Re-review on r6.

— Engineering Lead, 2026-07-25

o-ostrovskiy:
**Eng Lead — correcting a run-note miscommunication + re-affirming the r6 fix-list.** Reviewed @92b0191 (head unchanged since r5).

r5's country-qualified unify is correct and verified — the compose-path defect (stops filed under the wrong city at composition time) is genuinely fixed. That half is done.

But the 20:17 UTC note "no new fix-list, both findings → MYS-677, nothing to build" is **not right** — it conflated the two current-head Codex threads. Only **one** went to MYS-677:

- **[r3650496251](https://github.com/o-ostrovskiy/storyland-ai/pull/261#discussion_r3650496251)** — expansion stops bypass `_finalize` entirely (the whole guardrail chain) → **MYS-677**, post-launch, broader `executor.py` change. Correctly split out. Do NOT widen this PR to cover it.

- **[r3650496252](https://github.com/o-ostrovskiy/storyland-ai/pull/261#discussion_r3650496252)** — this is **BLOCKING for r6, and is NOT MYS-677.** When reconcile removes a city, `_finalize` returns the composer's suggestions unchanged. A chip whose `action_prompt` names the removed city can't match any persisted city, so `executor.expand` (checked at `executor.py:1157-1166`) falls back to `cities[0]` and persists the expansion under the **wrong city** — the exact defect this ticket exists to kill. **r5 creates this:** `main` never drops the city, so its chip never misroutes; r5's drop introduces a fresh wrong-city path one hop later (this class has now bitten this PR across r1–r5). Shipping r5 as-is would persist e.g. a Cartagena expansion under a removed-Aracataca chip, violating AC-2 ("never rendered under the wrong city").

**Minimal r6 (only required change):** have `reconcile_stop_city_grouping` expose the set of removed city names; in `_finalize` (which already holds the itinerary + suggestions together), drop or retarget any suggestion whose `action_prompt` names a removed city. Add a fixture: remove a city, assert no surviving suggestion references it.

No launch pressure now (PH postponed to September; this no longer needs the Jul 27 train), so take the time to get r6 right — but it is required before merge. Staying In Review, Role: Staff Engineer.

— Engineering Lead, 2026-07-25

o-ostrovskiy:
Eng Lead re-review of r6 (fix-list, keep In Review). The core guard through r5 is solid and verified against the diff — Colombia re-file, Salem/Massachusetts state-segment, Buffalo own-city-wins, and Paris-TX-vs-Paris-FR country-qualification are all correct. r6's suggestion-pruning is the right scope (it closes the `expand()`→`cities[0]` misroute I flagged), but the new helper has **3 valid open Codex P2s on this head** — the merge holds until they're closed:

1. **r3651340093 (MUST):** `_drop_suggestions_naming_removed_cities` uses `any(name in prompt_lower for name in removed_names)` — a naive substring test that over-drops. Removed "York" is a substring of surviving "New York", so a legitimate "New York" chip is dropped even though `expand()` would still resolve it; more broadly any removed name that's a substring of ordinary prompt text ("Bath"→"bathhouse", "Nice"→"nice restaurants", "Split") false-drops. The doc-comment claim *"same check expand() makes … survives iff expand() would still resolve it"* is **not true** — it's strictly stricter. Fix: drop a chip only if it names a removed city **and** does not still resolve to a surviving persisted city (mirror `expand()`'s actual first-match resolution), and correct the doc claim.

2. **r3651340088:** `_city_names` collapses to bare lowercased names, so a removed same-named / different-country city (London-UK emptied into London-Canada) is undetectable in the before/after diff — the exact MYS-548 "identity match blind to country" class this PR has already hit twice. Country-qualify the removed-city identity.

3. **r3651340087 (lowest):** `_address_city_candidates` strips a resolved trailing country unconditionally, so a city-state address ("Marina Bay, Singapore", "Central, Hong Kong") loses the only segment that names the CityPlan — a same-trip misfile there goes unreconciled. Preserve a resolved trailing segment as a locality candidate when it also slug-matches an itinerary city, while still using it for country qualification. This is fail-open under-reach, not a regression — fold into r7 if cheap, or split to a follow-up ticket if you want r7 minimal.

Not blocking, confirmed still correct to defer: r3650496251 (ExpansionResult bypasses `_finalize`) stays tracked as MYS-677, post-launch.

`mergeable_state: blocked` — re-run `make test-ci` on the runner (coverage floor). Ticket stays In Review, Role: Staff Engineer.

Reviewed @940850edd7c99aa1f5795f33790f7ef74b6e73e8 — Engineering Lead, 2026-07-25

o-ostrovskiy:
**Eng Lead merged 2026-07-26 — review-clean, CI green on the PR head, mergeable.** Squash `7b4179a9`. Reviewed @8feea38 (r8).

**Checked (10-pt):** correctness/scope ✓ (deterministic post-composition guard in the existing `_finalize` guardrail chain, fail-open, no scope creep); security ✓ (no auth/PII/network surface — reads two fields already on the held record, zero upstream, zero spend); edge cases ✓ (empty/malformed address, single-fragment, own-city attestation, same-named/country-qualified, empty-city keep-vs-drop); tests + **CI green on head** ✓ (`unit-tests`, `integration-tests`, `pip-audit`, `gitleaks`, Why/What/Docs/Verification all success; SE's real `make test-ci` = 913 passed, cov 83.77% — first real green in 8 revs now that google-adk installs in the runner); supply-chain ✓ (no dep change); grounding — n/a (no rec/agent path change); docs ✓ (`no impact`, correct); wiring/honest-claims — see below; repeat-defect — n/a.

**Codex — 3 fresh non-outdated findings on the r8 head, all verified VALID but NON-BLOCKING (fail-open under-reach, not active harm):**
- **P1 r3651906532** (state-name city can't be reconciled) · **P2 r3651906533** (2-letter state abbr → wrong country) · **P2 r3651906534** (city-agnostic chip over-drop).

r8 only ever *suppresses* a re-file (adds a `continue`), so it is provably **strictly safer than r7** — it cannot introduce a new misfile, only fail to fix one. The reported defect (distinct-name misfile: Cartagena stops under Aracataca) is fixed and stable; none of the three findings renders new wrong data or makes a false user-facing claim. This PR was on its **7th hop** (each revision's fix created the next edge one hop later) — rather than spin an 8th, the three residual under-reach edges are split to **[MYS-679](https://linear.app/mystoryland/issue/MYS-679)** (post-launch hardening, with a root-cause direction: positional locality-slot to end the ping-pong).

**Not closed by this merge:** AC-1 (measure mismatch rate across a sample) and AC-3 (re-verify the live `/i/colombian-caribbean-coast-CSuiMFRDYBsS` page serves no stop under the wrong city) need persisted-store / live-site access this build env lacks — owed post-deploy to QA. Merge ≠ AC met. Rides the Jul 27 train.

PR #262, opened 2026-07-25
fix(ai): make SessionStateAccessor read-only, persist book_metadata via append_event

o-ostrovskiy:
# fix(ai): make SessionStateAccessor read-only, persist book_metadata via append_event

## What

`SessionStateAccessor` exposed setters (`book_metadata`, `selected_regions`, `failed`) and a `clear_final_itinerary()` method that mutated `session.state` in place. ADK's session state does not persist an in-place mutation across a `get_session()` call — only `session_service.append_event(session, Event(actions=EventActions(state_delta={...})))` does. `core/executor.py`'s `discover()` and `local_atmosphere()` both called `state.book_metadata = ...`, which was a silent no-op against persisted state. This PR deletes the setters/method (making the accessor read-only) and fixes both call sites to persist via `append_event`.

## Why

ADR #11 documents the in-place-mutation trap and is the same class of bug as the `/status` FAILED-after-retry incident that motivated it. The dead write is harmless today only by accident — `book_title`/`author` reads fall back to the `BOOK_TITLE`/`AUTHOR` seed keys set at session creation, so nothing currently reads a *stale* value. But [MYS-168](https://linear.app/mystoryland/issue/MYS-168) (`USE_DATABASE=true` → `DatabaseSessionService`) cannot see an in-place dict write by construction — this is a precondition of that flip, not an optional cleanup. It also directly affects `/status`: `api/routes.py::_derive_job_status` and `get_status` both read `book_metadata` from persisted state, so a job that should report `JobStatus.DISCOVERING` was reporting `SEARCHING` for the entire discovery phase.

## How

- `core/session_state.py`: deleted the `book_metadata` setter, `selected_regions` setter, `failed` setter, and `clear_final_itinerary()`. Grepped the whole repo (code + tests) for callers outside `session_state.py`'s own test file: only `book_metadata`'s setter was ever actually invoked, at the two now-fixed `core/executor.py` call sites — the other three (`selected_regions` setter, `failed` setter, `clear_final_itinerary`) were dead code, presumably superseded by the `append_event`-based call sites that already exist elsewhere in `executor.py` for those same keys (`compose()`'s retry-clear block already sets `SELECTED_REGIONS`/`FINAL_ITINERARY`/`JOB_FAILED` correctly via `append_event`; `_mark_session_failed` already sets `JOB_FAILED` correctly the same way). Deleting them is the fix, not just documentation of intent: as long as a setter exists, the next writer reaches for `state.x = y` because it reads like it works. Updated the class docstring to state the read-only contract and point future writers at the `append_event` pattern instead.
- `core/executor.py`: `discover()` (metadata write before Phase 2) and `local_atmosphere()` (metadata write before the composition workflow) now build `Event(invocation_id="system", author="system", actions=EventActions(state_delta={SessionStateKeys.BOOK_METADATA: book_metadata.model_dump()}))` and `await self._session_service.append_event(session, metadata_event)` — the identical pattern already used a few lines away in the same file for `compose()`'s retry-clear block and `_mark_session_failed`.

## Scope decisions

- **`clear_final_itinerary`'s doc-comment mentioned a compose-retry use case, but it was never actually called anywhere** (confirmed by repo-wide grep) — the real retry-clear path already sets `FINAL_ITINERARY: None` via `append_event` in `compose()`'s existing guard block. Removing the dead method doesn't change behavior on that path.
- Not touched: any read-only property, `compose()`, `expand()`, `recommend_books()` — none of them use the deleted setters; their existing `state_delta`/`append_event` call sites were already correct and are the pattern this PR extends to the two remaining bad call sites.

## Docs

`no impact` — no architecture, endpoint, route, or config change; an internal API tightening (accessor becomes read-only) plus fixing two call sites to use the persistence mechanism the rest of the file already uses. `core/session_state.py`'s own docstring is the doc for this contract and was updated in this diff.

## Verification

**Environment note:** this build environment's Python is 3.10; `google-adk>=2.4.0` requires Python ≥3.11 to install, so the real test suite (which imports `google.adk`/`google.genai` transitively via `core/__init__.py` and `core/executor.py`) cannot be collected or run here — pre-existing, documented gap (see storyland-team-os memory "storyland-ai-deps-cant-install-in-sandbox"), not new to this PR.

What was verified directly, without ADK:
- `core/session_state.py` has zero ADK/google dependencies, so it was imported directly (via a stub `core` package in `sys.modules`, bypassing `core/__init__.py`'s eager `WorkflowExecutor` import) and every accessor behavior was run by hand: all existing reads unchanged, all three setters and `clear_final_itinerary` now raise `AttributeError`/are absent (`hasattr` false) — matching the four new regression tests added to `tests/unit/test_core.py`.
- `python3 -m py_compile` clean on all four changed/added files.
- Repo-wide grep confirming no caller of the three dead setters/method exists outside `session_state.py`'s own (now-updated) tests.
- Read `core/executor.py`'s existing correct `append_event`/`EventActions` call sites (the `compose()` retry-clear block, `_mark_session_failed`) and modeled both fixed call sites on the identical pattern — same `Event(invocation_id="system", author="system", ...)` shape, same `state_delta` dict shape, same immediate `await self._session_service.append_event(session, event)` call.
- Confirmed by reading the surrounding code that the `session`/`state` local variables I removed (the `SessionStateAccessor(session.state)` assignment that only existed to call the now-deleted setter) were not read again later in either `discover()`'s or `local_atmosphere()`'s body — `discover()` re-fetches a fresh `session`/`run_state` after `pump_events()` under a different variable name; `local_atmosphere()` reuses the same `session` variable later (`active_session = refreshed if refreshed is not None else session`), which I preserved by keeping the variable name and only replacing the setter call, not the `get_session()` line itself.
- **Could not literally execute** `tests/unit/test_session_state_persistence.py` (new) here — it drives `WorkflowExecutor.discover()`/`.local_atmosphere()` against a real `services.session_service.create_session_service(use_database=False)` (ADK's `InMemorySessionService`), monkeypatching the workflow-factory functions to raise immediately after the metadata write (so no live Gemini call is needed, modeled on the existing `TestExecutorCacheHit.test_cache_hit_skips_discovery_chain` pattern in `tests/unit/test_cache.py`), then re-fetches session state via a **second, independent** `get_session()` call and asserts `book_metadata` survived. This is the exact assertion shape the Eng Lead's tech plan calls for — a test reading back from the same in-memory object the code just wrote to would pass even with the bug, because the defect only shows up on an independent re-fetch.

**Please run the real suite on merge/CI** (`make test-ci`, coverage floor per `pyproject.toml`) — this environment could not, for the reason above, not because it was skipped. Reviewer focus: `tests/unit/test_session_state_persistence.py`'s two tests are new and the ones this PR's actual bug-fix claim rests on; the `test_core.py` accessor changes are lower-risk (pure-Python, already hand-verified above).

### DIFF
--- a/core/executor.py
+++ b/core/executor.py
@@ -439,12 +439,20 @@ class WorkflowExecutor:
 
             yield MetadataReady(metadata=book_metadata.model_dump())
 
-            # Store in session state
+            # Persist book metadata via append_event: in-place mutation of
+            # session.state does not survive get_session() (MYS-172) -- the
+            # accessor's setters were silent no-ops here.
             session = await self._session_service.get_session(
                 app_name=APP_NAME, user_id=user_id, session_id=job_id
             )
-            state = SessionStateAccessor(session.state)
-            state.book_metadata = book_metadata.model_dump()
+            metadata_event = Event(
+                invocation_id="system",
+                author="system",
+                actions=EventActions(
+                    state_delta={SessionStateKeys.BOOK_METADATA: book_metadata.model_dump()}
+                ),
+            )
+            await self._session_service.append_event(session, metadata_event)
 
             # Phase 2: Discovery
             yield ProgressEvent(
@@ -843,8 +851,16 @@ class WorkflowExecutor:
             session = await self._session_service.get_session(
                 app_name=APP_NAME, user_id=user_id, session_id=job_id
             )
-            state = SessionStateAccessor(session.state)
-            state.book_metadata = book_metadata.model_dump()
+            # Persist via append_event: in-place mutation of session.state
+            # does not survive get_session() (MYS-172).
+            metadata_event = Event(
+                invocation_id="system",
+                author="system",
+                actions=EventActions(
+                    state_delta={SessionStateKeys.BOOK_METADATA: book_metadata.model_dump()}
+                ),
+            )
+            await self._session_service.append_event(session, metadata_event)
 
             yield ProgressEvent(
                 phase=Phase.COMPOSITION,
--- a/core/session_state.py
+++ b/core/session_state.py
@@ -41,12 +41,21 @@ class SessionStateKeys:
 
 
 class SessionStateAccessor:
-    """Typed wrapper around session.state dict.
+    """Typed, READ-ONLY wrapper around session.state dict.
 
     Usage:
         state = SessionStateAccessor(session.state)
         metadata = state.book_metadata  # typed access, no magic strings
-        state.selected_regions = [...]  # typed write
+
+    MYS-172: this accessor is deliberately read-only. ADK's ``session.state``
+    does not persist an in-place mutation (``state[key] = value``) across a
+    ``get_session()`` call -- only ``session_service.append_event(session,
+    Event(actions=EventActions(state_delta={...})))`` does. A setter here
+    would look like a typed write and silently be a no-op against persisted
+    state, which is exactly the bug this ticket fixes (two call sites in
+    ``core/executor.py`` did precisely that). Write session state directly
+    via ``append_event`` at the call site, next to the ``session`` object
+    the write needs -- do not re-add a setter to this class.
     """
 
     def __init__(self, state: dict):
@@ -55,10 +64,6 @@ class SessionStateAccessor:
     @property
     def book_metadata(self) -> Optional[dict]:
         return self._state.get(SessionStateKeys.BOOK_METADATA)
-
-    @book_metadata.setter
-    def book_metadata(self, value: dict) -> None:
-        self._state[SessionStateKeys.BOOK_METADATA] = value
 
     @property
     def region_analysis(self) -> Optional[dict]:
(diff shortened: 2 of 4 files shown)


o-ostrovskiy:
✅ **Review clean — holding one cycle for Codex only.** Read in full: the read-only tightening is correct (deletes the three dead setters + `clear_final_itinerary`, only `book_metadata`'s setter had live callers — the two `executor.py` sites this fixes), both call sites now persist via the identical `append_event(Event(actions=EventActions(state_delta={BOOK_METADATA: ...})))` pattern already used by `compose()`'s retry-clear and `_mark_session_failed`, and the new `test_session_state_persistence.py` asserts on an **independent** second `get_session()` (the only shape that would fail with the bug present). Docs `no impact` is fair — ADR #11 already documents the in-place-mutation trap this enforces; the docstring carries the read-only contract. `mergeable_state: clean` (required checks green on head).

Not merging this run for one reason only: this is live agent logic on the hottest path (`discover` + `local_atmosphere`, every job) and **Codex hasn't posted its review yet** (both sibling PRs were reviewed within 3 min; this one is 40+ min old with nothing). On a hot path launch-eve I'd rather read Codex before merging, and there's no urgency — the bug is latent today (`book_title`/`author` fall back to the seed keys, nothing reads a stale value; it's a precondition for the MYS-168 DB flip + a cosmetic `/status` DISCOVERING-vs-SEARCHING fix). **Merge next run** if Codex is clean/👍 and head is unchanged.

Reviewed @f243fb843164a6db4ce216b9ff56acf1b289e703
— Engineering Lead, 2026-07-25

o-ostrovskiy:
Eng Lead merged 2026-07-25 (squash `b3779030`) — review-clean, CI green on the PR head (`make test-ci` ran the new independent-`get_session()` persistence test), mergeable. Accessor is now read-only; both `executor.py` call sites persist `book_metadata` via `append_event`.

Codex did not post a review on this PR (2.5h; both sibling PRs were reviewed within 3 min) — merged on my own complete review, since Codex is advisory and a clean/green/mergeable hot-path fix shouldn't be held indefinitely for a no-show.

Reviewed @f243fb843164a6db4ce216b9ff56acf1b289e703

Issue #25, opened 2026-01-28
Create book search component with autocomplete


o-ostrovskiy:
## Description
Build book search with Google Books autocomplete.

## Tasks
- [ ] Create SearchInput component
- [ ] Add debounced API calls
- [ ] Display book suggestions with covers
- [ ] Handle selection

**Sprint:** S13 (Aug) | **Hours:** 5

PR #265, opened 2026-07-26
fix(ai): atomic expand()/recommend_books() concurrency guard + merge re-validation + word-boundary city match

o-ostrovskiy:
## What

`expand()` and `recommend_books()` guarded against a second overlapping request with a check-then-set on a session-state flag (`expansion_in_progress` / `book_recs_in_progress`). Two overlapping requests for the same `job_id` could both read the flag `False` from their own independent session snapshot before either's write landed, so both proceeded. This PR makes that check-then-set atomic via a per-`job_id` `asyncio.Lock`, re-validates `expand()`'s merged itinerary against the `TripItinerary` schema before persisting it, and switches `expand()`'s target-city text match from a plain substring test to a word-boundary match.

## Why

Tech Radar's AI review (MYS-401, 2026-07-09, re-confirmed still open 2026-07-23) flagged three issues on the same code path:

1. **Double-charge / lost update.** A double-clicked suggestion chip (or an FE retry) could run `expand()` twice concurrently: both read `expansion_in_progress=False` before either's `append_event` committed, both ran a full agent workflow (billing Gemini twice), and whichever finished last silently overwrote the other's merge into `FINAL_ITINERARY` — `expansion_count` ends at a lost-update value instead of the correct total.
2. **Unvalidated merge.** The itinerary merge mutates raw dicts and persists straight to `FINAL_ITINERARY` with no schema check. A formatter quirk anywhere upstream (a stop missing a required field, a `parent_city` mismatch yielding an empty `cities` list) would silently persist a structurally-degraded itinerary that `/status` re-serves forever after.
3. **Substring city match.** Target-city selection used `city_name.lower() in action_prompt_lower`, so a trip city literally named "York" would match an `action_prompt` mentioning "New York" and merge into the wrong city.

The Engineering Lead's tech plan (2026-07-10, re-confirmed 2026-07-23) scoped all three as one additive `repo: ai` PR — no user-facing surface, no grounding/rec-output change, so no eval/G4-spend gate.

## How

- `core/executor.py`: added `WorkflowExecutor._session_locks` (a `defaultdict(asyncio.Lock)`) and `_get_session_lock(job_id)`. Both guards now do their check-then-set inside `async with self._get_session_lock(job_id):`, **re-fetching the session under the lock** rather than trusting the snapshot captured earlier in the function — the second caller's re-fetch happens strictly after the first caller's write is durable (`InMemorySessionService.get_session` returns a copy, never a live reference, so a stale snapshot genuinely can't see a concurrent write — only a fresh, serialized re-fetch can). In-process lock only: sufficient because this service runs a single uvicorn worker (no `--workers` in the `Dockerfile`); a multi-process deploy would need a durable compare-and-set instead (documented inline, ties back to MYS-168/MYS-169).
- `core/executor.py`: before `expand()` persists the merged itinerary, it now runs it through `validate_trip_itinerary` (the same helper `core/extraction.py` already uses elsewhere). On failure, the merge is **not** persisted (`FINAL_ITINERARY` stays whatever it already was) and the client gets a typed `MergeValidationError` instead of a silently-corrupted success.
- `core/executor.py`: new module-level `_matches_city_as_standalone_word(city_name, action_prompt)` — a regex word-boundary match that also rejects a match whose preceding word is capitalized (so "York" doesn't match inside "New York"). Documented as a heuristic, not a gazetteer, in its own docstring.
- `docs/ARCHITECTURE.md`: new ADR #25 + a Summary-table row.

## Docs

`docs/ARCHITECTURE.md` — ADR #25 added (new architecture decision: per-session locking + merge re-validation on this code path), plus the Key Architectural Patterns summary row.

## Verification

- New `tests/unit/test_executor_expansion_concurrency.py`:
  - `TestExpandConcurrencyGuardIsAtomic` forces two `expand()` calls on the same `job_id` to genuinely interleave (a session-service wrapper yields via `asyncio.sleep(0)` on both `get_session` and `append_event` — without a yield point between the read and the write, the two calls just run back-to-back on this single thread and the race is never exercised) and asserts exactly one completes while the other is rejected `ExpansionInProgress`, `expansion_count` ends at 1, and the new place is merged exactly once. **Confirmed red against the pre-fix guard** (`git stash` on `core/executor.py` only): both calls completed, doubling the merge. Reviewer: this is the test that matters most on this PR — worth reading closely.
  - `TestExpandMergeRevalidation` merges a well-formed new place into a seed itinerary that's missing a required `CityPlan` field (models a pre-existing formatter quirk) and asserts the merge is rejected and `FINAL_ITINERARY` is left untouched. Confirmed red against the pre-fix code (no validation existed at all — it silently merged and persisted the still-broken itinerary).
  - `TestCityWordBoundaryMatch` unit-tests the new helper directly (same style as `test_core.py`'s existing MYS-167 tests of the sibling `_resolve_trusted_action_prompt` helper): rejects "York" inside "New York", rejects "York" inside "Yorkshire" (the old substring bug), matches "York" as its own mention, matches at string start.
- Full suite: `make test-ci` → 923 passed, 7 pre-existing failures (confirmed identical on unmodified `core/executor.py` via `git stash` — sandbox Python 3.10 lacks the `ExceptionGroup` builtin CI's 3.12 has; not something to fix here). Coverage 88.03% (floor 74%).
- No dependency/lockfile change. No API/schema/wire-contract change (concurrency + validation only) → no eval, no G4-spend gate per the tech plan.
- Size: this diff (executor.py + docs + tests) runs over the informal ~300-line guideline, but the Engineering Lead's tech plan explicitly scoped all three findings as **one additive PR** — the tech plan's own decomposition call takes precedence, and splitting the atomic-guard fix from its regression test (or from the merge-revalidation fix that shares the same guard-adjacent code) would leave an unreviewable partial diff. Flagging per the size-cap policy rather than silently exceeding it.


o-ostrovskiy:
**Eng Lead review — r2, NOT review-clean (1 blocking item). Stays In Review.**

Good progress: both r1 findings are resolved and now show as outdated threads —
- ✅ **P1 city-match** — `_matches_city_as_standalone_word` now only rejects a match when the `"<preceding> <city>"` phrase is itself another trip city (passed via `trip_city_names`); "Explore Bath" / "Discover Barcelona" / "Victorian London" route correctly, with the cap-verb and cap-adjective regression tests I asked for. Correct.
- ✅ **P2 SSE envelope** — both under-lock re-fetches (expand + recommend_books) are wrapped and route a transient backend error through the client-safe `SessionError` pair; `TestExpandLockRefetchSessionError` asserts `SessionError` + `WorkflowComplete`. Correct.

**BLOCKING — Codex P1 (new, on `e4ba8bcb`, `executor.py:~1235`) is VALID — confirmed by my own read:**
The lock now makes the *flag* check-then-set atomic, and the realistic double-click interleaving is correctly rejected (second caller sees the durable `expansion_in_progress=True` under the lock). But the admitted caller then runs `body()` on the **stale pre-lock snapshot** — the outer `state`, `itinerary`, `expansion_count`, and `last_suggestions` captured before the lock. The whole reason to re-fetch under the lock is to get a fresh authoritative view; right now that fresh `lock_session`/`lock_state` is used only to read the flag and then discarded.

Scope note so this doesn't get over-built: the realistic double-click race is already closed by the durable flag — this is about making the guard correct *independent of scheduling* rather than "correct as long as no caller is parked between its outer lookup and the lock." Fix is bounded: **after admitting under the lock, adopt the under-lock re-fetch as the admitted request's authoritative snapshot** — re-derive `state` / `itinerary` / `expansion_count` / `last_suggestions` (and the recommend_books equivalents) from `lock_state`/`lock_session`, not the outer capture. Add a regression test that admits the second caller *after* the first has persisted + cleared the flag and asserts it merges onto the first's result (count → 2, no chip reuse), not the stale seed.

This completes the concurrency fix; I don't expect a further hop. Re-review next run on the new head.

Reviewed @e4ba8bcbaba86e833a862b15b7515e868ec84663 — checked: correctness/scope, both r1 fixes, CI (unit/integration/pip-audit/gitleaks/pr-body-sections all green on head), mergeable (clean), Codex (2 outdated resolved, 1 new P1 valid → above), docs (ADR #25 ✓).

o-ostrovskiy:
**Eng Lead review r3 — NOT review-clean. Fix-list below; stays In Review.** `Reviewed @4bda2d6e5d13b6f3f7e08d79fe0d162e392e9696`

Checked: correctness & scope · security (no authz/PII/injection surface change) · edge cases · tests + CI green on the PR head (unit/integration/pip-audit/gitleaks/pr-body-sections all ✅, coverage 87.95% over floor) · supply-chain (no dep/lock change) · grounding (no rec/agent output change → no eval, no G4) · Codex · wiring & honest claims · docs (ADR #25 present, §Docs accurate) · mergeability (`clean`, base.sha `706677b5` == main tip).

**r2 fix-list: CLEARED.** The admitted caller now adopts `lock_state`/`lock_session` in both `expand()` and `recommend_books()`, and `TestExpandAdmittedCallerAdoptsFreshSnapshot` is a real red-then-green regression test. That specific finding is closed.

Two new Codex findings on this head. **Both verified VALID against the diff — neither is a false positive.**

---

**1. P1 — the adoption happens *after* the admission checks, so the gap it closes reopens one line later.** (`core/executor.py:1287`, mirrored at 1646)

The `state = lock_state` assignment sits below the flag-rejection block, which is itself below `expand()`'s admission checks. Those checks — `action_id` ∈ `state.last_suggestions`, `expansion_count` vs `HARD_EXPANSION_CAP`, and the derivation of `trusted_action_prompt` — still run against the **stale outer capture**. Net effect on the exact scenario this ticket exists to fix:

- User double-clicks a chip. A completes, consumes `chip-1`, `expansion_count` → 1, flag durably clear.
- B parked pre-lock with the old snapshot. B's admission checks pass (its stale `last_suggestions` still contains `chip-1`, its stale count is 0). B acquires the lock, sees the flag clear, **is admitted**, then adopts the fresh state.
- B runs a full agent workflow on a chip that no longer exists → **the second Gemini charge is still billed.** If the fresh count were at `HARD_EXPANSION_CAP`, B bypasses the cap.

The lost-update half is genuinely fixed; the double-charge half — finding #1 in this PR's own "Why" — is not. And `TestExpandAdmittedCallerAdoptsFreshSnapshot` asserts a consumed `chip-1` **must be admitted** (`"the admitted second caller should complete"`), so the suite now pins the gap open as intended behavior. That assertion has to invert.

**Fix — and please make it the structural one, not a fourth patch.** The recurring shape across r1→r3 is that admission is decided against two different snapshots. Restructure so there is exactly one: acquire the lock first, re-fetch once, and run *all* admission validation — flag, `action_id` membership, cap, `trusted_action_prompt` derivation — against that single fresh snapshot, rejecting with the existing typed errors (`ExpansionInProgress` / invalid-action / cap-reached) when the refreshed state no longer permits the request. Then the pre-lock read is a pure optimisation and cannot decide anything. Mirror in `recommend_books()`.

**2. P2 — `(?![\w-])` misroutes hyphenated compounds to `cities[0]`.** (`core/executor.py:137`)

`Find Bath-based literary experiences` / `Explore London-themed cafés`: the trailing `-` fails the lookahead, the intended city is rejected, and `expand()` falls back to `cities[0]` — researching and persisting stops under the wrong city on any multi-city trip. The old substring match accepted these, so this is a routing **regression** introduced here, and it is the same `cities[0]`-misroute class MYS-660 spent eight revisions closing. Suggested shape, consistent with the r2 fix: allow a hyphen follower unless the hyphenated compound is itself another city in the same itinerary (`other_city_names` is already threaded in). Please cover both the hyphen-follower and hyphen-inside-another-city cases in `TestCityWordBoundaryMatch`.

---

No other blocking findings — the lock, the merge re-validation, ADR #25 and the SSE-envelope wrapping all read correctly. Re-review next run.

o-ostrovskiy:
## Eng Lead review r4 — one blocking item; routing to In Progress, not another in-place fix-list

**Reviewed @3b92a3ac6f85e18533538e8f00f6735d68959878** (was `4bda2d6e`; reviewed the r4 `506df89` + r5 `3b92a3ac` range).

### Cleared

**r4 item 1 — my r3 P1 is properly closed, and closed structurally.** You didn't patch the adoption point; you deleted the pre-lock session read entirely, acquired the lock on `job_id` alone, and made the single under-lock fetch the thing every admission decision runs against, in the original precedence order. There is no second snapshot left for the bug to live in — that's the fix I asked for rather than a fourth hop. Rewriting the regression test that had been *pinning the stale-chip admission as correct* was the other half, and you did it without being asked twice.

**r4 item 2 (hyphen boundary) and r5 item 1 (filter alignment) — both correct.** I read the `extraction.py` diff rather than the commit message: `_drop_suggestions_naming_removed_cities` condition (b) now calls the same `_matches_city_as_standalone_word` that `expand()` calls, so "this filter says the chip still resolves" and "`expand()` actually resolves it" can no longer disagree. Moving the helper into `extraction.py` was the right direction to avoid the circular import. Your reasoning for deliberately leaving condition (a) as a plain substring is also right — it only gates whether (b) runs, so over-firing there costs a check, not a wrong drop. Worth saying plainly: r5 item 1 is a defect **this PR created** by changing `expand()`'s matcher in r2, and you found it before it shipped.

### Blocking — the r5 lock-registry bound reopens this PR's own core guarantee

Codex flagged it at 04:37 (P2). I confirmed it by reading `_get_session_lock` / `_evict_stale_session_locks` at the head, not by taking the finding on trust — it is valid:

```python
lock = asyncio.Lock()
self._session_locks[job_id] = lock      # inserted FIRST
self._evict_stale_session_locks()       # then eviction runs
```

At the moment eviction runs, the just-inserted lock is neither `locked()` nor has `_waiters` — the caller has not entered `async with` yet. The loop walks LRU→MRU skipping held/waited entries, so if every older entry is skipped it reaches the new entry and deletes it. The caller then holds a `Lock` that is **not in the registry**, and the next caller for that same `job_id` mints a *different* `Lock` — both run the admission path concurrently. That is precisely the TOCTOU double-charge/lost-update this ticket exists to close, reintroduced by the hardening added for it.

To be fair about severity: reaching it needs ~10,000 simultaneously held-or-waited locks, so this is not a live-traffic risk today. I'm not holding on likelihood — I'm holding because the serialization guarantee *is* this PR's deliverable, and a registry that can hand out an unregistered lock cannot be the thing that guarantees it.

**Fix — evict before inserting, not after:**

```python
lock = self._session_locks.get(job_id)
if lock is None:
    self._evict_stale_session_locks()   # make room FIRST
    lock = self._session_locks[job_id] = asyncio.Lock()
else:
    self._session_locks.move_to_end(job_id)
return lock
```

with the cap test adjusted to `>=` since you're now making room for one. This is structurally better than Codex's two suggestions (exclude-the-returned-key, or allow-temporary-overflow): the new entry is never an eviction candidate because it doesn't exist yet, so there's no exclusion parameter to forget and no second code path. If nothing is evictable, the registry simply sits one over cap for a moment — which is the correct trade, since a bounded-memory nicety must never win over the lock identity.

Please add the regression that pins it: fill the registry to cap with every entry **held**, request a genuinely new `job_id`, and assert the returned object `is` `_session_locks[job_id]` — i.e. that what the caller got back is what the next caller will find.

### Two process notes, and they matter more than the code above

**1. This PR is at r5 on one shape, so it goes back to In Progress rather than sitting In Review on a sixth fix-list.** Not a criticism of the work — r3, r4 and r5 each cleared the thing I actually asked for. But the pattern across all five rounds is the same: each fix creates the next finding one hop away (r2's matcher change → r5's filter drift; r5's registry bound → this lock-identity hole). Five rounds of me fix-listing and you patching is the loop itself misbehaving, not either of us. Own the whole guard as one piece and come back when it's coherent.

**2. Please stop adding surface to this PR.** The three findings in the ticket are done. The bounded registry was scope you took on because Codex raised it — legitimate, since the registry is new in *this* diff and shipping an unbounded one would be a real leak, so finish it. But if the next head draws a new *adjacent* Codex idea, **file it as its own ticket and tell me** rather than patching it in. Each addition resets the review surface, and that's what has kept this diff open through five rounds.

**Also:** `main` has moved (ai#266 merged as `fd42a04b`). No file overlap with your branch, but merge `origin/main` **into** this branch — a merge commit, never a rebase/force-push — before pushing r6.

Routing to **In Progress · Role: Staff Engineer** on [MYS-401](https://linear.app/mystoryland/issue/MYS-401).

o-ostrovskiy:
## Eng Lead review r6 — review-clean. **MERGED** as `0834128a`.

**Reviewed @dae84b2a22d40c337644abd76308115c0e0f0f1b** (reviewed the r6 range: `ded15ff` main-merge + `dae84b2a`; r1–r5 reviewed on prior heads).

### The bounce is closed, and closed the way I asked

I read `_get_session_lock` / `_evict_stale_session_locks` at the head rather than taking the commit message on trust:

```python
lock = self._session_locks.get(job_id)
if lock is not None:
    self._session_locks.move_to_end(job_id)
    return lock
self._evict_stale_session_locks()
lock = asyncio.Lock()
self._session_locks[job_id] = lock
return lock
```

Both return paths are now safe **by construction, not by ordering luck**:

- **Existing entry** → MRU-touch and return; eviction never runs on this path, so the lock being handed back cannot be a candidate.
- **New entry** → eviction runs *before* the insert, so the only entries it can iterate are ones that existed before this call. The entry the method is about to return does not exist yet.

`_evict_stale_session_locks` re-checks `len(...) < CAP` inside the loop (so it stops as soon as there's room for one), skips `lock.locked() or getattr(lock, "_waiters", None)`, and deletes LRU-first. An empty `_waiters` deque correctly reads falsy — an entry with no pending waiter is evictable, which is what we want. The pathological "nothing evictable → registry sits one over cap" branch is documented as an accepted trade, and it is the right trade: a memory nicety must never win over lock identity.

The regression test pins the exact scenario (cap=1, one entry held, brand-new `job_id` → the returned lock **is** `_session_locks[job_id]`, and a second call returns the identical instance). That's the assertion that matters — it tests registry *identity*, not just registry *size*.

You also honoured the scope ask: 44 added lines in `executor.py`, no adjacent hardening smuggled in, and you merged `origin/main` in rather than force-pushing.

### Full checklist on this head

Correctness & scope ✓ · security (no authz/PII/injection/secret surface) ✓ · edge cases (cap boundary, held/waited skip, all-entries-held) ✓ · **tests + CI green on the PR head** — all 5 required checks ✅ on `dae84b2a` (unit-tests, integration-tests, pip-audit, gitleaks, Require Why/What/Docs/Verification), 955 passed, coverage 90% over the 74% floor · supply-chain (no dep or lockfile change; pip-audit clean) ✓ · grounding (concurrency + routing only, no rec/agent output change → no eval, no G4-spend) ✓ · wiring & honest claims (no data-sending control; ADR #25's claims match the code I read) ✓ · **docs track code** — `## Docs` present, ADR #25 rewritten to describe the eviction ordering as the design rather than a patch ✓ · mergeability `clean`, `base.sha` `fd42a04b` == `main` tip at merge time ✓.

**Codex:** last review is on the r5 head (`3b92a3ac`, 04:37Z); it has not commented on `dae84b2a` 50+ minutes after the push, against a ~3-minute latency on every prior revision — read as no suggestions. The two threads still showing not-outdated are the two I verified closed myself (r5's `extraction.py` filter alignment, and this eviction hole). Codex went 6-for-6 on valid findings across this PR — worth recording as the reason it gets read carefully rather than skimmed.

### Two things worth saying out loud

1. **Six revisions, and the last one was the good one.** r1→r5 all had the same shape — each fix opened the next finding one hop away. r6 is different in kind: you re-owned the guard instead of patching hop six, and the invariant is now stated in the docstring and pinned by a test. That's the standard.
2. **This does NOT ride today's train.** The Jul 27 RC was cut Sunday 17:05 pinning `ai` at `706677b5`; this merged after, same as ai#266. Flagged on the ticket so "Merged" isn't misread as "shipping today".

Eng Lead merged 2026-07-27 — review-clean, CI green on the PR head, mergeable.

PR #266, opened 2026-07-26
test(ai): AI auth-boundary test gaps — cross-user session isolation, inflight-slot lifetime over a real stream + on disconnect, sweeper resolver-store wiring

o-ostrovskiy:
## What

Test-only PR closing three of the four gaps the Tech Radar AI review flagged: cross-user session isolation, inflight-slot lifetime over a real stream (both the normal-completion and mid-stream-disconnect cases), and the session sweeper's coverage of the lazily-created place→book resolver store. No production code changes.

## Why

The AI suite is genuinely behavioral, but the auth boundary — the weakest link — was the least-tested. Three concrete gaps:

1. **No negative `X-Internal-Secret` test** — already closed by MYS-399's own PR (`tests/unit/test_gateway_auth.py` now has full missing/wrong/empty-config coverage). Verified present; no new work needed for this gap.
2. **No cross-user / session-isolation test** — every existing status/compose/expand test ran as the implicit `dev_user`, so a regression that dropped or mis-threaded `user_id` anywhere on the path would have shipped unnoticed.
3. **Inflight-slot lifetime was only ever driven as a bare async generator** — never through a real streaming response, so "released when the SSE generator finishes OR the client disconnects mid-stream" was unverified. Also untested: whether the session sweeper's provider actually includes the lazily-created place→book resolver store once it exists.

## How

- `tests/unit/test_session_isolation.py` (new): creates a real session as user `alice` (real `InMemorySessionService`, same technique as `test_executor_expansion_concurrency.py`), then proves user `bob` cannot read it — a genuine HTTP 404 via `api/routes.py::get_status`, and an SSE `JobNotFound` (never a leaked itinerary) via `core.executor.WorkflowExecutor.compose()`/`expand()`.
- `tests/unit/test_api.py`: new `TestInflightLimiterOverRealStream` drives `limit_inflight()` through a genuine ASGI round-trip (httpx `AsyncClient` + `ASGITransport`, same pattern the file's existing streaming-endpoint tests use) and asserts the slot frees after the stream completes — and that the freed slot is genuinely reusable, not just zeroed by accident. New `TestSessionSweeperIncludesResolverStoreOnceCreated` drives the real `api/app.py::lifespan` context manager (with `initialize()`/`shutdown()` stubbed, `SessionSweeper` itself real) and proves `_live_session_services` picks up the resolver's session store once it's lazily created.
- `tests/unit/test_ratelimit.py`: new `TestLimitInflightCancellation` covers the mid-stream-disconnect half. I initially tried this over a real ASGI round-trip (abandon a `client.stream()` read early), but confirmed empirically that httpx's `ASGITransport` in this environment does not propagate an early stream-read abort as a genuine server-side cancellation — it just hangs for the full generator lifetime instead of disconnecting. So this is tested at the dependency-generator level instead: the task holding the slot open is cancelled mid-`sleep`, and its cleanup calls `gen.aclose()` — the same unwind mechanism Starlette uses on a real client disconnect — and the test asserts the slot is freed.

## Docs

No impact — test-only change, no architecture/contract/config/route change.

## Verification

- New tests: `tests/unit/test_session_isolation.py`, `tests/unit/test_api.py::TestInflightLimiterOverRealStream` + `::TestSessionSweeperIncludesResolverStoreOnceCreated`, `tests/unit/test_ratelimit.py::TestLimitInflightCancellation`.
- All new tests confirmed passing individually with `--timeout=15` (pytest-timeout) to rule out any hang; the disconnect scenario was validated to genuinely fail (timeout) before the ASGI-transport approach was replaced with the cancellation-based one — worth a close read since it's the trickiest test in the set.
- Full suite: `make test-ci` → 936 passed, 7 pre-existing failures (identical shape/count to the current `main` baseline — sandbox Python 3.10 lacks the `ExceptionGroup` builtin CI's 3.12 has, unrelated to this change). Coverage 88.96% (floor 74%).
- No dependency/lockfile change, no API/schema/wire-contract change → no eval, no G4-spend gate.

### DIFF
--- a/tests/unit/test_api.py
+++ b/tests/unit/test_api.py
@@ -4,6 +4,7 @@ Unit tests for FastAPI SSE API layer.
 Tests API models, endpoint responses, and SSE streaming with mocked executor.
 """
 
+import asyncio
 import json
 
 import pytest
@@ -1438,9 +1439,19 @@ class TestPlaceToBookEndpoint:
 
 class TestInflightLimiterOverRealStream:
     @pytest.mark.asyncio
-    async def test_slot_is_released_after_a_completed_stream(
+    async def test_slot_is_held_during_the_stream_and_released_after(
         self, mock_app_state, mock_executor
     ):
+        """MYS-403 (Codex P2): the original version of this test fully
+        buffered the request (a plain ``client.post``) before inspecting
+        ``active`` -- so a regression that freed the slot as soon as the
+        endpoint FUNCTION returned, rather than when the SSE generator
+        actually finishes draining, would still have passed (it ends at 0
+        either way). This drives a genuine partial ASGI read: the mock
+        generator parks mid-stream (after its first event, before its
+        last) until the test explicitly releases it, so ``active`` is
+        inspected while the response is demonstrably still open.
+        """
         from api.app import create_app
         from api.ratelimit import InFlightLimiter
         from httpx import AsyncClient, ASGITransport
@@ -1448,10 +1459,16 @@ class TestInflightLimiterOverRealStream:
 
         mock_app_state.inflight_limiter = InFlightLimiter(max_in_flight=1)
 
+        hold = asyncio.Event()
+        first_event_sent = asyncio.Event()
+
         async def mock_discover(**kwargs):
             yield ProgressEvent(
                 phase=Phase.BOOK_SEARCH, step="Searching Google Books API"
             )
+            first_event_sent.set()
+            # Parked mid-stream until the test releases it below.
+            await hold.wait()
             yield WorkflowComplete(job_id="test-job")
 
         mock_executor.discover = mock_discover
@@ -1461,20 +1478,50 @@ class TestInflightLimiterOverRealStream:
         deps._app_state = mock_app_state
 
         client = AsyncClient(transport=ASGITransport(app=app), base_url="http://test")
-        response = await client.post(
+
+        async def _drive_stream():
+            async with client.stream(
+                "POST",
+                "/api/v1/itinerary/discover",
+                json={"book_title": "1984", "author": "George Orwell"},
+            ) as response:
+                assert response.status_code == 200
+                async for _ in response.aiter_lines():
+                    if first_event_sent.is_set():
+                        break
+
+        stream_task = asyncio.create_task(_drive_stream())
+        await asyncio.wait_for(first_event_sent.wait(), timeout=5)
+        # Give the event loop a few ticks so the generator has genuinely
+        # parked on `hold.wait()` (not just scheduled to) before checking.
+        for _ in range(5):
+            await asyncio.sleep(0)
+
+        assert mock_app_state.inflight_limiter.active == 1, (
+            "the slot must still be held WHILE the SSE generator is "
+            "mid-stream, not only before the endpoint function returns"
+        )
+        # At capacity 1, a concurrent request must be shed while the first
+        # is still open -- proves the slot is genuinely occupied, not just
+        # a counter that happens to read 1.
+        concurrent = await client.post(
             "/api/v1/itinerary/discover",
             json={"book_title": "1984", "author": "George Orwell"},
         )
-        assert response.status_code == 200
+        assert concurrent.status_code == 503
+
+        hold.set()
+        await asyncio.wait_for(stream_task, timeout=5)
+
         assert mock_app_state.inflight_limiter.active == 0, (
             "the slot must be free once the SSE generator has fully drained"
         )
         # And the freed slot is genuinely usable, not just zeroed by accident.
-        second = await client.post(
+        after = await client.post(
             "/api/v1/itinerary/discover",
             json={"book_title": "1984", "author": "George Orwell"},
         )
-        assert second.status_code == 200
+        assert after.status_code == 200
 
 
 # =============================================================================
--- a/tests/unit/test_api.py
+++ b/tests/unit/test_api.py
@@ -1412,3 +1412,132 @@ class TestPlaceToBookEndpoint:
     async def test_missing_place_rejected(self, test_client, reset_p2b_singleton):
         response = await test_client.post("/api/v1/place-to-book", json={})
         assert response.status_code == 422
+
+
+# =============================================================================
+# MYS-403 gap 3a: inflight-slot lifetime over a REAL ASGI stream
+# =============================================================================
+#
+# api/dependencies.py::limit_inflight was previously exercised only as a bare
+# async generator (tests/unit/test_ratelimit.py) -- never through a real
+# streaming response, so its own docstring's claim ("holds a slot for the
+# duration of the request ... releases it when the response completes") was
+# unverified. A leak here silently wedges capacity at MAX_INFLIGHT_REQUESTS
+# forever, shedding every subsequent request with 503 until a restart.
+#
+# The completed-stream half lives here (a genuine ASGI round-trip). The
+# mid-stream-disconnect half lives in test_ratelimit.py::
+# TestLimitInflightCancellation -- httpx's ASGITransport in this environment
+# does not propagate an early `async with client.stream(...)` exit as a real
+# server-side cancellation (confirmed empirically: it hangs for the full
+# generator lifetime instead), so that half is tested at the dependency-
+# generator level by cancelling the task holding the slot open, which is
+# exactly the mechanism Starlette uses to unwind a dependency generator on a
+# genuine client disconnect.
+
+
+class TestInflightLimiterOverRealStream:
+    @pytest.mark.asyncio
+    async def test_slot_is_released_after_a_completed_stream(
+        self, mock_app_state, mock_executor
+    ):
+        from api.app import create_app
+        from api.ratelimit import InFlightLimiter
+        from httpx import AsyncClient, ASGITransport
+        import api.dependencies as deps
+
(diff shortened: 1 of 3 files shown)


o-ostrovskiy:
**Eng Lead review (test-only, MYS-403) — NOT review-clean. Stays In Review.**

Net-positive PR — cross-user isolation coverage is genuinely new and worth having. Two things before merge:

**Mechanical (done for you):** the required `Require Why/What/Docs/Verification` check was failing because the description used H1 (`#`) section headings; the gate wants H2 (`##`). I've edited the description to `##` — that gate should go green on the next run. Use `##` for What/Why/How/Docs/Verification going forward.

**3 Codex P2s — all valid; for a test-only PR the assertions *are* the deliverable, so these are blocking:**

1. **`test_api.py` `TestInflightLimiterOverRealStream` (slot held *during* the stream):** the request is fully buffered before `active` is inspected, so a regression that frees the slot as soon as the endpoint returns — before the SSE generator drains — would still pass (ends at 0, reusable). Block the mock generator mid-stream, assert `active == 1` (or a concurrent request gets 503) *while* it's open, then let it finish.

2. **`test_ratelimit.py` `TestLimitInflightCancellation` (disconnect via the real path):** the test's own `finally` calls `gen.aclose()`, so it proves *your* cleanup runs, not that Starlette actually unwinds `limit_inflight` on a client disconnect — it'd pass even if production cleanup were broken. Drive a raw ASGI harness (construct the scope + a `receive` that yields `{"type": "http.disconnect"}` while the response generator is parked) rather than httpx — that covers the integration behavior the test claims to protect. If a real `http.disconnect` genuinely can't be simulated even at the raw-ASGI level, say so in a comment and I'll accept it with justification, but httpx being the blocker isn't sufficient — raw ASGI is the standard approach here.

3. **`test_session_isolation.py` (isolation through the route stack) — the important one:** calling `executor.compose()`/`expand()` directly only proves the executor's session lookup is user-scoped. This PR's stated purpose is protecting the *auth boundary* — the HTTP surface — and that's exactly where a regression would hide: a route or streaming adapter that stops forwarding the authenticated `user_id` and falls back to the shared default would leave these tests green while HTTP callers cross the boundary. Exercise the endpoints with Alice's real session and a Bob identity (header/override) so route→stream→executor propagation is covered. The `get_status` test already does the route-level version — extend the same pattern to compose/expand.

Reviewed @4cad6c6f461f7019ea03b4e6b3c0d760113a53e4 — checked: scope (test-only, no prod code ✓), CI (unit/integration/pip-audit/gitleaks green; body-sections was red → fixed above), mergeable, Codex (3 valid P2 → above), docs (test-only, no impact ✓).

o-ostrovskiy:
**Eng Lead review r2 — one finding left; stays In Review.** `Reviewed @2a6bce7b8728a0c3cb739c379e3c7d68568318cb`

Checked: spec/assertion correctness · security (auth-boundary surface — the point of the PR) · edge cases · CI green on the PR head (unit/integration/pip-audit/gitleaks/pr-body-sections all ✅, coverage 87.38% over floor) · supply-chain (no dep/lock change) · Codex · honest claims · docs (`## Docs — no impact` accurate for a test-only diff) · mergeability (`clean`, base.sha `706677b5` == main tip).

**Two of three prior items: CLEARED, and cleared properly.**

1. **Mid-stream slot hold** — the generator now parks on `hold.wait()` after its first event, `active == 1` is asserted while the stream is demonstrably open, and a concurrent request sheds `503` at capacity 1. That last assertion is what makes it a real test rather than a counter read — good.
2. **Raw-ASGI disconnect** — `test_release_runs_on_a_real_http_disconnect_over_raw_asgi` drives a hand-built scope/receive/send trio against the real app and delivers a genuine `http.disconnect`, so Starlette's own unwind is now what's under test rather than the harness's own `aclose()`. This is exactly the right harness, and it's reusable — see below.

**3. Route-level isolation — partially closed; one gap remains (Codex P2, `test_session_isolation.py:242`). Valid.**

`TestComposeRouteIsolation` / `TestExpandRouteIsolation` do close the real gap I flagged: route → `compose_stream`/`expand_stream` → executor propagation, confirmed red by hardcoding the forwarded `user_id`. But they call the route functions with `user_id=` supplied as an already-resolved argument, so `Depends(get_gateway_user_id)` never runs — no header parsing, no dependency resolution. A regression that left the parameter in place but repointed or dropped its `Depends` (falling back to a shared default identity) keeps every test in this file green while HTTP callers cross the boundary.

That matters more than a normal P2 here because this is a test-only PR: the assertions **are** the deliverable, and the PR body claims it closes the auth-boundary gap. As written the suite doesn't yet prove the identity resolver is wired to these routes at all — a claim ahead of what the code demonstrates, which is the class my checklist blocks on.

**Fix (small, and the machinery already exists in this PR):** add one test that drives the real ASGI app end-to-end for alice's session — either the raw-ASGI harness you just built in `test_ratelimit.py`, or `AsyncClient`/`ASGITransport` with the gateway identity header set — asserting bob gets `JobNotFound` and alice does not, with identity arriving through `get_gateway_user_id` rather than as a parameter. One compose case is enough; expand is the same wiring. Keep the existing route-level tests, they're still worth having.

This is the last open item — nothing else on the diff blocks. Re-review next run.

o-ostrovskiy:
## Eng Lead review r3 — APPROVED, merging

**Reviewed @0e0223ac35f0135b18e56e2d36a47a90f7123c79** (was `2a6bce7b`; reviewed the r3 `57f8d5e` + r4 `0e0223ac` range).

### My r2 fix-list item is CLEARED, and cleared at the mechanism level
The remaining item was: *the route-level isolation tests pass `user_id=` in as an already-resolved arg, so `Depends(get_gateway_user_id)` never runs — a dependency-wiring regression leaves the file green.*

r3 closes it properly rather than cosmetically. `TestRouteIsolationOverRealGatewayIdentity.real_identity_client` does `app.dependency_overrides.pop(get_gateway_user_id, None)` on its own `create_app()` instance and drives a genuine `ASGITransport` round-trip with a real `X-User-ID` header — so the identity dependency under test is the real one, not conftest's autouse `"test_user"` stand-in. I checked that line specifically: it is the load-bearing part of the whole claim, and it's there. Keeping the direct-call classes alongside it is right — they still pin route→stream→executor propagation, which the ASGI tests don't isolate.

Two things I want to credit explicitly, because they are the difference between a test and a claim:
- The **deliberate regression check** (hardcoding `compose()`'s `user_id` to alice → the bob test failed → reverted). That is the only evidence that makes a green auth-boundary test mean anything.
- Catching the **CRLF parsing bug** — `EventSourceResponse` writes `\r\n`, so splitting on `"\n\n"` collapsed the body into one block and only the last event survived. An early version of this test passed *for the wrong reason*. Self-caught false-green; exactly the class we keep getting burned by.

### r4's two Codex findings — both valid, both genuinely closed
1. **Positive expand over a real header** — correct catch. A route that resolved *every* expand request to a fixed non-owner identity would still emit `JobNotFound` for bob and pass. `test_alice_header_reaches_her_own_session_via_expand` closes the asymmetry.
2. **Status route still direct-call** — same bypass class, just not extended to `get_status`. The new real-ASGI alice/bob pair asserts `200` + `book_title == "Persuasion"` on the owner and `404` + no `"Bath"`/`"Persuasion"` in the body on the non-owner. The negative test asserting the *payload* doesn't leak, not just the status code, is the right shape.

Your call to leave the three older Codex threads' lines unchanged and close their concern via sibling tests is correct and matches my r2 instruction — I verified the sibling coverage exists rather than taking the commit message for it.

### Checklist
1. **Correctness & scope** ✅ test-only; maps to MYS-403's gaps 2–4 (gap 1 verified already closed by MYS-399's own PR, not re-done).
2. **UI** — n/a.
3. **Security** ✅ strengthens the auth boundary's coverage. `verify_gateway_secret` staying overridden is justified and justified *in writing* — that boundary has dedicated negative coverage in `test_gateway_auth.py`.
4. **Edge cases** ✅ positive + negative identity per route (compose / expand / status); slot held mid-stream, freed after drain, freed on real disconnect, 503 at capacity.
5. **Tests + CI green on the PR head** ✅ `unit-tests`, `integration-tests`, `pip-audit`, `gitleaks`, `Require Why/What/Docs/Verification` all success on `0e0223ac`. Read the PR's own checks, not the local run.
6. **Supply-chain** ✅ no dependency or lockfile change; `pip-audit` green.
7. **Grounding** — n/a, no rec/agent logic touched.
8. **Codex** ✅ 6 threads across 4 heads; every one either fixed in-place or closed by a sibling test I verified myself.
9. **Wiring & honest claims** ✅ for a test-only PR the assertions *are* the deliverable, so I read them rather than the summary. They assert real behavior through the real dependency. `## Docs — no impact` is a true claim for this diff.
10. **Docs track code** ✅ `## Docs` present and accurate.

`mergeable_state: clean`; base.sha `706677b5` == ai `main` tip, so `clean` is meaningful and not a stale-base artifact.

### One follow-up, not a merge blocker
The real-ASGI identity class now covers compose / expand / status. Any *other* route carrying `Depends(get_gateway_user_id)` is still only covered at the direct-call level. That's outside MYS-403's flagged scope and I'm not holding a 4th revision for it — but if a future PR adds an identity-guarded route, its real-ASGI ± pair belongs in this class rather than a new file.

**Eng Lead merged 2026-07-27 — review-clean, CI green on the PR head, mergeable.**

PR #267, opened 2026-07-27
test(ai): dead-field inventory class guard — fail CI when a prompt reads a preference field no UI collects

o-ostrovskiy:
## What
Adds `tests/unit/test_dead_field_inventory.py`: an AST scanner over the prompt-assembly surface (`agents/*.py`, `core/prompts.py`) that collects every literal field name read off a variable/parameter named `preferences` via `preferences["field"]` or `preferences.get("field")`, and fails CI if that name isn't documented in the paired inventory file `docs/ui_collected_preference_fields.json` (which names the FE surface that collects it) — or if the inventory documents a field the scanner no longer finds anywhere (stale-entry drift, checked in both directions).

## Why
MYS-436 deleted `reader_profile_agent`, the only prompt that ever read a named preference field (`budget`, `preferred_pace`, etc.), so there's no live violation today — every remaining prompt builder (`core/prompts.py::build_composition_prompt`, `agents/prompts.py::preferences_block`) treats `preferences` as an opaque dict, passed through whole via `json.dumps(preferences)`. That doesn't make the guard pointless: per the ticket's own PM ruling, this is a class-level recurrence guard ("second occurrence = write the gate"), not a re-proof of an already-fixed instance — its job is catching the *next* prompt wired to a field the UI never collects, which is the actual failure mode MYS-392 hit (the mood-chip's `preferences.moods`, read by nothing).

## How
- `find_named_preference_reads()` walks each scanned file's AST looking for `ast.Subscript`/`ast.Call` nodes whose receiver `Name` is literally `preferences`, collecting string-literal keys. Deliberately AST- not regex-based (this is Python, `ast` is stdlib and immune to the false-positive/line-break issues a regex scanner has to work around).
- Scope is intentionally narrow and documented as such in the module docstring: bound to the literal receiver name `preferences`, over `agents/` + `core/prompts.py` only — no alias/destructuring tracking, no attribute-style access (the unused `TravelPreferences` model). Verified by hand today that no such alias exists in the scanned surface. Mirrors the same narrow-the-claim move made on the fe sibling guard (`storyland-web/tests/regionCitiesGuard.test.ts`, MYS-681, built earlier in this run) rather than over-promising "every read."
- `docs/ui_collected_preference_fields.json`: `fields` map (empty today) + a `_readme` explaining why it's empty and how to add an entry. QA's weekly dead-control sweep can diff against the same file.
- Self-check test pins every offense shape (subscript, `.get()`, `.get()` with a default, multiple reads) and every safe shape already shipped in this repo (`json.dumps(preferences)`, `dict(preferences)`, `.items()`, a differently-named receiver) so the detector itself is covered, not just the production-tree scan.
- **r2 (Eng Lead fix-list, 2 valid Codex P2s):** `visit_Subscript()` didn't check `node.ctx`, so `preferences["derived"] = value` (Store) and `del preferences["derived"]` (Del) were both counted as reads — the service writing/removing its own key, not a field the FE must have collected. Gated on `isinstance(node.ctx, ast.Load)`; added store-context + `del` self-check cases. Separately, `test_inventory_file_is_valid_json_with_a_fields_map()` only checked `fields` was a dict — `{"budget": null}` or `{"budget": ""}` passed while the `_readme` and failure message both promise each entry names the FE surface that collects it; now asserts every value is a non-empty string.

## Docs
`docs/ui_collected_preference_fields.json` *is* the doc for this guard (own `_readme` explains scope, rationale, and how to extend it). No `ARCHITECTURE.md` entry: this is a test-only addition with no architecture, contract, endpoint, env, or route change.

## Verification
- `python3 -m pytest tests/unit/test_dead_field_inventory.py -v` — 3/3 pass.
- Full suite: `python3 -m pytest tests/unit/ --cov --cov-report=term-missing` — 958 passed, 7 pre-existing failures (all `ExceptionGroup`-dependent, `test_discovery_errors.py`/`test_run_harness.py`, Python-3.11-only builtin, confirmed identical on `main` with these two new files absent — untouched by this PR). Coverage 89.90% (floor 74%). `ruff check` clean on the touched file.
- To verify the guard actually catches something: temporarily add `preferences.get("budget")` anywhere in `agents/` or `core/prompts.py` and re-run — the inventory test goes red naming the undocumented field; add a `fields` entry and it goes green again.
- To verify the r2 write/del fix: temporarily remove the `ast.Load` gate and add `preferences["derived"] = 1` anywhere in the scanned surface — pre-fix, the self-check's new store-context case fails (falsely collected as a read); post-fix it passes.
- Environment note: this sandbox's `requirements-dev.lock` pins `rpds-py==2026.6.3`, which requires Python ≥3.11 and has no py3.10 wheel/sdist; the sandbox only has 3.10. Installed the full dep set via `pip install --no-deps -r ` + a compatible standalone `rpds-py==0.30.0`, confirmed `google.genai` imports and the full suite collects/runs correctly afterward — a one-time environment workaround, not a project change (lockfile untouched, not part of this diff).

### DIFF
--- a/docs/ui_collected_preference_fields.json
+++ b/docs/ui_collected_preference_fields.json
@@ -1,0 +1,21 @@
+{
+  "_readme": [
+    "MYS-439 class-level guard inventory (paired with tests/unit/test_dead_field_inventory.py).",
+    "Every preference FIELD NAME literally read off a variable/parameter named `preferences` inside",
+    "the prompt-assembly surface (agents/*.py, core/prompts.py) via preferences[\"field\"] or",
+    "preferences.get(\"field\") must have an entry in `fields` below, naming the FE surface that",
+    "actually collects it -- so QA's weekly dead-control sweep can diff against the same list. The",
+    "test fails CI if a scanned field is missing here, or if an entry here names a field the scan no",
+    "longer finds (stale documentation of a read that was removed).",
+    "",
+    "Empty today (2026-07-27): MYS-436 deleted reader_profile_agent, the only prompt that ever read",
+    "a NAMED preference field. Every remaining prompt builder treats `preferences` as an opaque dict",
+    "passed through whole via json.dumps(preferences) -- core/prompts.py::build_composition_prompt,",
+    "agents/prompts.py::preferences_block. The FE does not collect or send any named preference",
+    "field on the live discover/compose paths either (storyland-web's `preferences` payload is",
+    "optional and unused since MYS-392 retired the mood-chip channel). Add an entry here in the SAME",
+    "PR that adds a named preferences.get(\"field\")/preferences[\"field\"] read to the prompt-assembly",
+    "surface."
+  ],
+  "fields": {}
+}
--- a/tests/unit/test_dead_field_inventory.py
+++ b/tests/unit/test_dead_field_inventory.py
@@ -1,0 +1,186 @@
+"""
+MYS-439 — class-level guard: fail CI when a prompt reads a preference field
+no UI collects.
+
+## Why this exists (class, not instance)
+
+The class this test guards against is: a prompt-assembly module reads a
+NAMED preference field (`preferences["budget"]`, `preferences.get("pace")`)
+that nothing on the frontend actually collects and sends — the model then
+receives (and can act on, or narrate) a field that is always absent. This
+ticket was originally scoped against `reader_profile_agent`'s prompt, which
+named five such fields. MYS-436 deleted that agent outright, so there is no
+live violation today; this test passes cleanly on the current tree.
+
+That does not make the guard pointless (PM ruling, 2026-07-27, MYS-439):
+this is a "second occurrence = write the gate" class-level guard (the
+sibling read-side of MYS-437's send-side wired-control spec), and its value
+is catching the *next* time someone wires a prompt to a field the UI never
+collects — not re-proving today's already-fixed instance. Instance fixed !=
+class closed.
+
+## What counts as an offense
+
+Every prompt-assembly module (`agents/*.py`, `core/prompts.py`) currently
+treats `preferences` as an OPAQUE dict, passed through whole via
+`json.dumps(preferences)` — `core/prompts.py::build_composition_prompt`,
+`agents/prompts.py::preferences_block`. Neither ever reads a field by name.
+An offense is a NEW literal-key read off a variable/parameter named exactly
+`preferences`: `preferences["field"]` or `preferences.get("field", ...)`.
+Reading the whole dict opaquely (`json.dumps(preferences)`,
+`preferences.items()`, `dict(preferences)`) is not an offense — that's the
+safe, already-shipped shape.
+
+## SCOPE (deliberately narrow, not exhaustive)
+
+This is an AST scanner bound to the literal receiver name `preferences`,
+over a fixed set of prompt-assembly files (`agents/` + `core/prompts.py`).
+It does NOT track a differently-named alias (`user_preferences`,
+`prefs = preferences; prefs["budget"]`), attribute-style access on a typed
+object (`preferences.budget`, e.g. if `TravelPreferences` — currently
+unused in production code — were wired back in), or reads outside the
+scanned files. Verified by hand as of this PR that no such alias exists
+today. Mirrors the same "receiver-name-bound, narrow the claim rather than
+over-promise" scope note the fe sibling guard carries
+(`storyland-web/tests/regionCitiesGuard.test.ts`, MYS-681).
+
+## Inventory file
+
+`docs/ui_collected_preference_fields.json`: `fields` maps each read field
+name to the FE surface that collects it. Empty today (see the file's own
+`_readme`). A field found by the scanner but missing from the inventory
+fails CI; a field listed in the inventory but no longer found by the
+scanner also fails (stale documentation of a read that was removed —
+catches the inventory drifting from reality in either direction).
+"""
+
+import ast
+import json
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+INVENTORY_PATH = REPO_ROOT / "docs" / "ui_collected_preference_fields.json"
+
+# The prompt-assembly surface. NOT core/executor.py, core/session_state.py,
+# core/cache.py, etc: those pass the whole `preferences` dict through
+# opaquely (`.items()` for cache-key hashing, `state["user:preferences"] =
+# preferences`) without ever reading a NAMED field — that's a different
+# (and already-guarded, MYS-401) class, not this ticket's.
+SCAN_ROOTS = [REPO_ROOT / "agents", REPO_ROOT / "core" / "prompts.py"]
+
+
+def _iter_py_files():
+    for root in SCAN_ROOTS:
+        if root.is_file():
+            yield root
+        elif root.is_dir():
+            yield from sorted(root.rglob("*.py"))
+
+
+class _PreferencesFieldVisitor(ast.NodeVisitor):
+    """Collects every literal field name read off a variable/parameter
+    literally named `preferences` via `preferences["x"]` or
+    `preferences.get("x", ...)`.
+    """
+
+    def __init__(self):
+        self.fields: set[str] = set()
+
+    def visit_Subscript(self, node: ast.Subscript) -> None:
+        if isinstance(node.value, ast.Name) and node.value.id == "preferences":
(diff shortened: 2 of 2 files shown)


o-ostrovskiy:
## Eng Lead review — Reviewed @3a15e92101067955efb5f76e7e29b0f856b58c60

Good guard, right shape: AST over regex is the correct call in Python, the scope note is honest about what it doesn't track, and the bidirectional (undocumented **and** stale) check is more than the ticket asked for. Three fix-list items, all small.

### 🔴 1. CI is RED — this cannot merge as-is

`Require Why/What/Docs/Verification` **failed**, and `mergeable_state: blocked`. The body has `## What` / `## Why` / `## How` / `## Docs` but the last section is titled **`## Review & test`**; the required check wants a literal **`## Verification`** heading. Rename it. (Content is fine — this is a heading, not a gap.) Everything else is green: `unit-tests`, `integration-tests`, `pip-audit`, `gitleaks`, `Seer Code Review`.

### 🔴 2. Codex P2 (valid): write contexts counted as reads

`visit_Subscript` doesn't inspect `node.ctx`, so `preferences["derived"] = value` and `del preferences["derived"]` are both collected as reads. The guard would then demand an FE-collection entry for a key the service **creates internally** — a false red that a future author can only silence by documenting a frontend surface that doesn't exist, which quietly corrupts the inventory QA diffs against. Gate on `isinstance(node.ctx, ast.Load)` and add a store-context + `del` case to the self-check.

### 🔴 3. Codex P2 (valid): the inventory's stated contract isn't enforced

`test_inventory_file_is_valid_json_with_a_fields_map` checks only that `fields` is a dict; `{"budget": null}` or `{"budget": ""}` passes every assertion. But the `_readme` and the failure message both promise each entry **names the FE surface that collects the field** — so the guard can go green on an inventory that tells QA nothing. That's this repo's own "copy claims something the code doesn't do" class ([MYS-492](https://linear.app/mystoryland/issue/MYS-492)/[MYS-584](https://linear.app/mystoryland/issue/MYS-584)), and it's the *whole* payload of the entry. Assert each value is a non-empty string (or narrow the `_readme`'s claim to match — but non-empty is the cheaper half).

### Checked and fine

- **Scope**: test + docs only, 2 new files, zero production code touched. No prod-behavior risk, so nothing here is security-, migration- or grounding-relevant.
- **Honest claims**: the SCOPE docstring correctly declares the receiver-name binding, the fixed scan roots, and the untracked alias/attribute forms — and the test name matches the narrowed claim. The "empty today and here's why" `_readme` is the right way to ship a guard with no live violation; **instance fixed ≠ class closed** is exactly the PM ruling and I agree with building it.
- **Repeat-defect rule**: this *is* the class gate, so it satisfies its own bar.
- **Deps/lockfile**: no dependency change in the diff. The `rpds-py` py3.10 workaround is correctly confined to the sandbox with the lockfile untouched — right call, and `pip-audit` is green.
- **§Docs**: accepting "the JSON file is the doc" for a test-only addition. No `ARCHITECTURE.md` impact.
- **7 pre-existing suite failures**: not a blocker — `unit-tests` is green on this head, which is the check that governs.
- **Self-check quality**: pinning the safe passthrough shapes (`json.dumps`, `dict()`, `.items()`, differently-named receiver) alongside the offense shapes is what makes this a detector test rather than a detector claim. Keep that when fixing (2).

Issue stays **In Review**, `Role: Staff Engineer`. [MYS-439](https://linear.app/mystoryland/issue/MYS-439)

o-ostrovskiy:
## Eng Lead review r2 — one item left, and I own why it exists

**Reviewed @bb114d42.** CI is green now (`Require Why/What/Docs/Verification` ✅ on the re-run at 16:09:49, unit ✅, integration ✅, pip-audit ✅, gitleaks ✅), `mergeable_state: clean`, `base.sha 0834128a` == ai main tip. Both r1 Codex P2s are correctly fixed and the self-check cases you added for them are real (store-context and `del` fixtures, not label pins). Test-only diff, no prod code, no dep or lockfile change.

### The one blocker: my own r1 fix-list opened a hole, and Codex's newest P2 is right

Gating on `ast.Load` excludes writes — which is what I asked for and it is correct — but it also excludes the one Store-context shape that **semantically reads the prior value**: an augmented assignment. `preferences["budget"] += suffix` evaluates the existing FE-supplied value before storing, and its target `Subscript` carries `ast.Store`, so the scanner now returns nothing and that named read ships with no inventory entry.

Direction is **false-GREEN** — CI says "no uninventoried reads" when it can't see one. That's the MYS-492/584 class the guard's own `_readme` promises to close, so it isn't shippable as a disclosed gap.

### Why this is the LAST round, not round 4 of whack-a-mole — I proved the set is closed

I did not want to hand you another "widen it again" after what fe#295 cost us, so I enumerated it against the actual grammar instead of arguing from intuition. A `Subscript` node can carry exactly three contexts — `Load`, `Store`, `Del`; there is no fourth. I generated every syntactic position that produces a non-`Load` `Subscript` and checked which one reads the prior value:

| position | ctx | reads prior value |
|---|---|---|
| `Assign` target | Store | no |
| `AnnAssign` target | Store | no |
| **`AugAssign` target** | **Store** | **YES** |
| `For` / `AsyncFor` target | Store | no |
| `With` optional_vars | Store | no |
| comprehension target | Store | no |
| tuple / starred unpack | Store | no |
| `Delete` | Del | no |

So **read-set = `Load` ∪ (`Store` where the parent is `AugAssign`)**, and that is a fixed point. This is exactly the difference I wrote up on MYS-685: a source-text regex has no enumerable node set and each round finds a narrower variant, whereas `ast` does — which is why this guard is the right mechanism and gets to be *finished* rather than perpetually widened.

**The fix:** add `visit_AugAssign` (or check the parent when the target is a `preferences` subscript), collect the literal key, and pin a self-check case for `preferences["budget"] += x` alongside the existing store/`del` safe cases. ~5 lines.

**Do NOT widen anywhere else while you're in there.** `preferences.pop("x")` and `.setdefault("x", …)` also read, but they sit outside the module docstring's *stated* offense definition (`preferences["field"]` and `preferences.get("field")`), so they are disclosed scope, not a false claim. `AugAssign` is different in kind: it IS `preferences["field"]`, so the code currently contradicts its own documented contract. Closing that contradiction is the whole ask. If you think `.pop`/`.setdefault` are worth covering, that's a separate ticket and a scope-note edit, not this round.

Everything else stands — AST over regex is the right call in Python, the SCOPE note is honest about receiver-name binding, and the bidirectional stale-entry check exceeds what the ticket asked for.

o-ostrovskiy:
## Walkthrough — ready for your review & merge

**What & why.** A CI test that fails the build if an AI prompt ever starts reading a user-preference field (`preferences["budget"]`) that no part of the website actually collects. When that happens the model silently receives a field that is always empty, and can narrate around it — that's what MYS-392 did with the mood chip. There is no such bug today; this is the gate that catches the next one.

**How it fits.** Test-only, `storyland-ai`. It parses the prompt-building files into Python's syntax tree and compares every named field read against a checked-in inventory file (`docs/ui_collected_preference_fields.json`, empty today) that names which screen collects each field. Drift in either direction fails: an unlisted read, or a listed field nobody reads any more.

**Trade-offs.** The obvious alternative was a text search for `preferences["`. We tried that shape on the frontend twin (MYS-681) and it took three rounds of widening as new spellings slipped past. A syntax-tree scanner has a *finite, enumerable* set of shapes to cover, which is why this one converged. That difference is the whole content of MYS-685.

**Worth asking.** (1) The scan is bound to variables literally named `preferences` — rename one to `prefs` and the guard goes quiet. That limit is written into the module docstring rather than hidden; ask whether a guard that documents its own blind spot is the right trade against one that over-promises. (2) `.pop()` / `.setdefault()` also read a field and are deliberately excluded — I told the engineer not to widen to them, because they sit outside the offense definition the docstring states. Disclosed scope, not a false claim.

**Review notes.** CI 5/5 green on this head (unit, integration, pip-audit, gitleaks, PR-template). `mergeable_state: clean`. Codex raised three findings across three rounds; all three were valid and all three are fixed — the last one (`preferences["x"] += y` reads the old value even though Python marks it a write) was a false-green my own r1 instruction introduced. I proved the read-set is closed before asking for this round: `Load` ∪ (`Store` under `AugAssign`), and of every syntax position that yields a non-`Load` context, only `AugAssign` reads first. No G3/G4 implication — test-only, no spend, no prod surface.

Ready for your review & merge — Reviewed @22033ca6ee8f0d5b2482298a4cde6ef1d1522f13

PR #268, opened 2026-07-27
feat: GET /api/v1/version reports the deployed commit SHA (MYS-687)

o-ostrovskiy:
## Why
Confirming "is the latest release actually on prod" currently means SSH + `docker inspect` + timestamp cross-referencing. Olga asked directly for a one-request way to check (MYS-687).

## What
New `GET /api/v1/version` reusing the existing `SENTRY_RELEASE` env var (already stamped by `deploy.sh` at deploy time for the Sentry SDK) — a second consumer of the same value, no new env var.

## Docs
No `ARCHITECTURE.md` impact — additive endpoint, mirrors the existing `/health` shape, no new architecture/contract/route pattern.

## Verification
`TestVersionEndpoint` in `tests/unit/test_api.py` — configured-SHA case and empty-fallback case. Full suite: 966/966 passing (964 pre-existing + 2 new).

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -5,6 +5,8 @@ Maps HTTP endpoints to streaming generators and status queries.
 All business logic lives in core.executor — routes are thin wiring.
 """
 
+import os
+
 from fastapi import APIRouter, Depends, HTTPException
 from sse_starlette.sse import EventSourceResponse
 
@@ -79,6 +81,14 @@ async def health_check() -> HealthResponse:
         )
     except RuntimeError:
         return HealthResponse(status="unhealthy", version="0.1.0")
+
+
+@system_router.get("/version")
+async def version() -> dict:
+    """Deployed commit SHA — reuses SENTRY_RELEASE (already stamped by
+    deploy.sh at deploy time; same value the Sentry SDK tags releases with),
+    so "what's live" is one request instead of an SSH session."""
+    return {"service": "storyland-ai", "git_sha": os.environ.get("SENTRY_RELEASE") or "unknown"}
 
 
 @router.post(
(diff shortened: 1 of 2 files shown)


o-ostrovskiy:
## Walkthrough — ready for your review & merge (3 of 4 for MYS-687)

The ai leg of the `/version` work. Merge alongside [storyland-infrastructure#57](https://github.com/o-ostrovskiy/storyland-infrastructure/pull/57), [storyland-services#123](https://github.com/o-ostrovskiy/storyland-services/pull/123) and [storyland-web#297](https://github.com/o-ostrovskiy/storyland-web/pull/297). Full walkthrough for the set is on infra#57.

**What & why.** `GET /api/v1/version` returning `{"service": "storyland-ai", "git_sha": …}`.

**How it fits.** This one needs no new plumbing at all — the ai container has been given `SENTRY_RELEASE` at deploy time for months so Sentry can tag releases. This is a second reader of the same value. That's why the ai leg is four lines while the backend's needs a config field.

**Trade-offs.** It reads `os.environ` at request time rather than going through the config object the way the backend does. Slightly inconsistent between the two services; the upside is it can't drift from what Sentry is actually reporting, which is the point of reusing the variable.

**Worth asking.** The three services report different `service` names — `storyland-ai`, `backend`, `frontend`. That looks like sloppiness and isn't: each matches its own docker-compose service name, so what you curl and what you'd `docker compose restart` line up. Worth deciding if you'd rather they matched the repo names instead, while there are no consumers yet.

**Review notes.** CI 8/8 green on this head (unit, integration, pip-audit, gitleaks, PR-template, Seer). Two tests covering set and unset. No Codex findings. No G3/G4 implication — additive read-only endpoint, no spend, no data path.

Ready for your review & merge — Reviewed @01cb6caa6af1afa00dc915dfcc846e329b4a2adc

PR #277, opened 2026-08-02
fix(sentry): drop bare-query span/trace-data keys the '?' scrub misses (MYS-551 ai twin)

o-ostrovskiy:
## Why

Twin of [MYS-551](https://linear.app/mystoryland/issue/MYS-551) confirmed 2026-08-02 (the Eng Lead's infra#59 review scoped the finding to the gateway and flagged this service as an unchecked surface; checking it found the same `if "?" in value` predicate on span data). `span.data['http.query']` holds the bare query with no leading `?` — the scrub never fires, and with tracing on, user search terms (`book_title=…`, `place=…`) would survive a scrubbed transaction. Gates the `.env.prod` `SENTRY_TRACES_SAMPLE_RATE=1.0` flip (MYS-735).

## What

Same fix as the gateway ([services#127](https://github.com/o-ostrovskiy/storyland-services/pull/127)) and frontend ([web#310](https://github.com/o-ostrovskiy/storyland-web/pull/310)): `_scrub_data_record` deletes query-named keys (`(^|[._])query($|[._])`, case-insensitive) and keeps the `?`-tail scrub for other strings — applied to every span's `data` and the root span's `contexts.trace.data`.

## Docs

Inline docstrings carry the MYS-551 rationale and cross-repo parity references. No standalone docs page covers scrubbing internals.

## Verification

**Red-before-green proven**: both new tests fail against pre-fix `main`, pass after. Suite: 1004 passing; the 4 failing integration tests (`test_cache_real_api`, `test_langfuse_integration` ×3) **fail identically on clean main** — environment-dependent (live API / Langfuse env), untouched by this diff.

### DIFF
--- a/api/sentry.py
+++ b/api/sentry.py
@@ -82,6 +82,27 @@ logger = get_logger(__name__)
 # in access lines and any other URL-bearing stdlib log reaching Sentry Logs.
 _URL_QUERY_RE = re.compile(r"\?[^\s\"]*")
 
+# Span/trace attribute keys whose VALUE is a bare query string with no
+# leading '?' (MYS-551, ai twin of the gateway defect): sentry-sdk's HTTPX
+# integration stores span.data['http.query'] = 'q=<user search>'. The
+# '?'-heuristic below can never fire on those values, so the KEY is dropped
+# outright. Matches http.query / url.query / query_string style names.
+_QUERY_KEY_RE = re.compile(r"(^|[._])query($|[._])", re.IGNORECASE)
+
+
+def _scrub_data_record(data):
+    """
+    Scrub one span/trace attribute dict in place (MYS-551): query-named keys
+    are DELETED (their whole value is the query — nothing to strip back to),
+    every other string value loses any '?' tail. Same two rules as the
+    gateway (storyland-services#127) and the frontend (storyland-web#310).
+    """
+    for key, value in list(data.items()):
+        if _QUERY_KEY_RE.search(key):
+            del data[key]
+        elif isinstance(value, str) and "?" in value:
+            data[key] = _URL_QUERY_RE.sub("", value)
+
 
 def _drop_health_probe_logs(log, hint):
     """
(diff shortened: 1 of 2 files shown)


o-ostrovskiy:
## Walkthrough — ready for your review & merge

**What & why.** The ai service's Sentry span scrubber only cleaned values containing a `?`. The SDK stores a request's query string in `span.data['http.query']` as a bare `book_title=…` with no `?`, so the scrubber never fired on the one field that is nothing but user input. Inert while tracing is off; a live export of reader search terms the moment it's on.

**How it fits.** This is the twin of [services#127](https://github.com/o-ostrovskiy/storyland-services/pull/127) — the ai service and the gateway carry near-identical `sentry.py` scrubbers by design, and this defect was in both. Worth naming how it was found: I scoped [MYS-551](https://linear.app/mystoryland/issue/MYS-551) to the gateway and flagged this service as *unchecked, not clean*; someone then checked it and it had the same hole. **The unchecked-surface note is what produced this PR** — that's the note doing its job rather than being a hedge.

**Trade-offs.** Two rules instead of one: query-named keys are **deleted**, everything else keeps the `?`-tail strip. Deleting rather than blanking is deliberate — the whole value is the query, so there's nothing to strip back to. The alternative was a stricter value-shaped heuristic (does this look like `k=v`?), which would false-positive on legitimate attributes. Here the failure direction is safe: over-matching loses a debug attribute, under-matching leaks a reader's search.

**Worth asking.**
- The tests feed synthetic events straight into `_scrub_event`. That proves *"if the SDK sends `http.query`, we drop it"* — not *"the SDK sends `http.query`"*. The second half is inherited from a July Codex finding. It's the right way to test this, and worth knowing which half is pinned.
- The regex needs a delimiter (`.`/`_`), so a camelCase `searchQuery` wouldn't match. No SDK we run emits that shape — but it's a name enumeration, and enumerations are what got us here.

**Review notes.** CI 🟢 **5/5** on `40198b18`, sticky SHA-matched — Unit, Integration, pip-audit, Gitleaks, PR body. ⚠️ The PR body flags 4 integration tests failing locally as pre-existing; **CI's Integration Tests ran green**, so that was a sandbox-environment artefact and is a non-issue for merge. `mergeable_state: clean`, base current. I verified the registration rather than the diff alone — `_scrub_event` is wired to **both** `before_send` and `before_send_transaction`, so the span path is live; on `before_send` alone every line of it would be dead code that these tests still pass. No dependency change. Docs §6 honest. No G3/G4.

⚠️ **Sequencing, so a green merge isn't over-read:** this and services#127 are a pair. Merging this one alone does **not** make server-side tracing safe — the gateway's twin is still red on CI. The `.env.prod` flip waits for both.

Ready for your review & merge — Reviewed @40198b18

— Engineering Lead, 2026-08-02

